PR #1, opened 2015-01-13
fix, array length has to be of type size_t

BBasile:
...otherwise build fails on a 32 bit OS

### DIFF
--- a/src/dfmt.d
+++ b/src/dfmt.d
@@ -53,7 +53,7 @@ int main(string[] args)
     else
     {
         File f = File(args[1]);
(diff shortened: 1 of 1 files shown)


src/dfmt.d
```diff
@@ -53,7 +53,7 @@ int main(string[] args)
     else
     {
         File f = File(args[1]);
-        buffer = new ubyte[](f.size);
+        buffer = new ubyte[](cast(size_t)f.size);
```
Hackerpilot: I always forget this.

PR #3, opened 2015-01-14
More stuff

qznc:
The commits makefile, space-after-return, and noDoubleNewline are quite trivial.

The comment-commit is more complex. Look into tests/comments.d.ref to get an impression of the results.

I also added some tests I have, which currently fail. Some things are obvious bugs, but with others I am not sure what the desired behavior actually is.

### DIFF
--- a/src/dfmt.d
+++ b/src/dfmt.d
@@ -115,8 +115,25 @@ private:
         assert (index < tokens.length);
         if (current.type == tok!"comment")
         {
+            const i = index;
+            if (i > 0)
+            {
+                if (tokens[i-1].line < tokens[i].line)
+                {
+                    if (tokens[i-1].type != tok!"comment"
+                        && tokens[i-1].type != tok!"{")
+                        newline();
+                }
+                else
+                    write(" ");
+            }
             writeToken();
-            newline();
+            if (i >= tokens.length-1)
+                newline();
+            else if (tokens[i+1].line > tokens[i].line)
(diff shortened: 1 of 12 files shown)


src/dfmt.d
```diff
@@ -115,8 +115,25 @@ private:
...
+                    output.write(" ");
+            }
             writeToken();
-            newline();
+            if (i >= tokens.length-1)
+                newline();
+            else if (tokens[i+1].line > tokens[i].line)
+                newline();
+            else if (tokens[i+1].type != tok!"{")
+                output.write(" ");
```
Hackerpilot: output.write doesn't increase the counter for the length of the current line the way that write() does. Is this intentional?

tests/guessnumber.d.ref
```diff
@@ -0,0 +1,36 @@
...
+        {
+            writeln("  Out of range!");
+            continue;
+        }
+        if (answer == target)
+        {
+            writeln("  Well guessed.");
+            break;
+        }
+        writeln(answer < target ? "  Too low.":"  Too high.");
```
Hackerpilot: At some point I need to fix this, probably by asking the AST where the attribute declarations and labels are.

src/dfmt.d
qznc: No. Will fix it.

qznc:
output.write => write.

Issue #7, opened 2015-01-16
Heronian 


qznc:
Here is a code snippet, which is really hard to format. No idea, how to approach it. It is taken from [Rosetta Code: Heronian Triangles](http://rosettacode.org/wiki/Heronian_triangles#D).

```
const h = cartesianProduct(r, r, r)
          //.filter!({a, b, c} => ...
          .filter!(t => t[0] <= t[1] && t[1] <= t[2] &&
                        t[0] + t[1] > t[2] &&
                        t[].gcd3 == 1 && t[].isHeronian)
          .array
          .schwartzSort!(t => tuple(t[].hero, t[].only.sum, t.reverse))
          .release;
```

Let me try to break this into sub-problems:
- we indent the comment wrong, because the temporary indent of expressions is calculated at the "."
- we do not support indenting to relevant expression above, instead of vertically aligning t[0], we indent by four spaces
- to find good line breaks in long expressions like the filter function, we probably need some weighting on expression depth. We want to break lines at &&, not at + or >.
- we need a non-greedy approach to break the line after ".array". Currently, we continue ".array.schwartz..." and break within "t.reverse".

Hackerpilot:
In this case using ```// dfmt off``` and ```// dfmt on``` is the best solution.

MartinNowak:
> we need a non-greedy approach to break the line after ".array". Currently, we continue ".array.schwartz..." and break within "t.reverse".

How about either putting all dot chains on a single line or, if that's too long, each on it's own line.
Reading https://google.github.io/styleguide/cppguide.html#Function_Call,  which is a very similar problem, suggests that we'd need to preserve/understand existing formatting decisions.

I guess there are more important/easier to implement issues than this one.

Hackerpilot:
I'm convinced that the code formatter that people actually *want* is `echo`.

WebFreak001:
fixed with #471 using `--keep_line_breaks true`

(only a small line break bug with the comment which is a separate issue)

PR #8, opened 2015-01-16
Bunch of improvements

qznc:
* No trailing space after for,foreach,etc
* Put some space before @
* More clever length estimating for import statements …
* Fixing a bug in token length calculation

### DIFF
--- a/src/dfmt.d
+++ b/src/dfmt.d
@@ -197,6 +197,28 @@ private:
                     newline();
                     break;
                 }
+                else if (current.type == tok!",")
+                {
+                    // compute length until next , or ;
+                    int length_of_next_chunk = INVALID_TOKEN_LENGTH;
+                    for (size_t i=index+1; i<tokens.length; i++)
+                    {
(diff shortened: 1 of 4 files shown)


src/dfmt.d
```diff
@@ -197,6 +197,28 @@ private:
...
+                    // compute length until next , or ;
+                    int length_of_next_chunk = INVALID_TOKEN_LENGTH;
+                    for (size_t i=index+1; i<tokens.length; i++)
+                    {
+                        if (tokens[i].type == tok!"," || tokens[i].type == tok!";")
+                            break;
+                        const len = tokenLength(i);
+                        assert (len >= 0);
+                        length_of_next_chunk += len;
+                    }
```
qznc: This whole approach could be generalized into some clever algorithm which *rather* breaks according to precedences.  Also see issue #7

Hackerpilot: Yes. I've been thinking about the best way to do this.

PR #10, opened 2015-01-17
Allow for outputting to anything supporting the write() method. Make it ...

jcd:
...possible to exclude main.

These changes makes it possible for other programs to compile and link with dfmt directly.

### DIFF
--- a/src/dfmt.d
+++ b/src/dfmt.d
@@ -75,7 +75,7 @@ int main(string[] args)
             else
                 break;
         }
-        format("stdin", buffer, output);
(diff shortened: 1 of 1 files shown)


Hackerpilot:
Did you intend for this to be usable with output ranges? Right now Files support ```write``` and ranges use ```put```. It may be better to standardize on ```put``` and supply the formatting code with ```stdout.lockingTextWriter()```.

jcd:
I can change to output range if you want. I tried to do the least intrusive change that would work for my use case in the hope that it was easier to get through a pull request.

PR #15, opened 2015-02-22
Add pesky brace handling

kuettler:
Get those pesky semicolons and curly braces out of my code. Idea from
https://twitter.com/thedirtycoder/status/569339014085517312 . Not exactly useful,
but quite impressive. I hope you like it.

### DIFF
--- a/src/dfmt.d
+++ b/src/dfmt.d
@@ -112,6 +112,7 @@ Formats D code.
     --tabs | -t      Use tabs instead of spaces for indentation
     --braces=allman  Use Allman indent style (default)
     --braces=otbs    Use the One True Brace Style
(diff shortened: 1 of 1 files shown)


kuettler:
This is a hack, really. I spend just enough time on it to make it work. See `./bin/dfmt --braces=pesky src/dfmt.d |less` for the desired outcome. ;)

Hackerpilot:
I'm not sure if I should be impressed or horrified.

kuettler:
I am fine with both. :) If you do not want to include it, fair enough. It's a joke. On the other hand, it is actually fun looking at the code that way. So it might be worthwhile. Did you give it a try?

PR #175, opened 2015-08-20
enum can also be used in is() expressions

### LINKED ISSUE #174: Too much space
Yet another dog food issue. In editorconfig.d this time.

```d
void merge()
{
    static if (is(T == enum))
        *thisN = x;
}
```
should be unchanged, but becomes
```d
void merge()
{
    static if (is(T == enum ))
            *thisN = x;
}
```
Unwanted space after enum and too much indentation for if-body.

qznc:
fixes #174

I guess is() expressions have more problems, though. I'm not sure how to approach this systematically.

My check for == and ) might not be enough either.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -794,12 +794,16 @@ private:
                 write(" ");
             break;
         case tok!"enum":
-            if (peekBackIs(tok!"identifier"))
-                write(" ");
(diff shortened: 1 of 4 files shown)


Hackerpilot:
Both of the .ref files are here, but it looks like you forgot to check in the file that should be formatted.

qznc:
Added missing test source file

Issue #176, opened 2015-08-20
Eat your own dog food


qznc:
Applied to itself with default configuration, dfmt source code should not change. More precisely:
```sh
dfmt --inplace src/
```
You should see no change in "git diff" which was not there before.

There are also various changes, which I would consider correct. In a few places we probably want to disable formatting explicitly.

This is kind of a meta issue, since #172 #173 #174 are actual issues that come up, if you do this and maybe there will be more.

PR #180, opened 2015-08-25
Eat your own dogfood

qznc:
Indented to tackle #176 but I do not expect this to be merged right away. However, we can have a discussion here, what maybe should be changed.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -32,8 +32,8 @@ void format(OutputRange)(string source_desc, ubyte[] buffer, OutputRange output,
     astInformation.cleanup();
     auto tokens = byToken(buffer, config, &cache).array();
     auto depths = generateDepthInfo(tokens);
-    auto tokenFormatter = TokenFormatter!OutputRange(buffer, tokens, depths, output,
-        &astInformation, formatterConfig);
+    auto tokenFormatter = TokenFormatter!OutputRange(buffer, tokens, depths,
+        output, &astInformation, formatterConfig);
     tokenFormatter.format();
 }
 
@@ -75,7 +75,7 @@ struct TokenFormatter(OutputRange)
      *         decisions.
      */
     this(const ubyte[] rawSource, const(Token)[] tokens, immutable short[] depths,
-    OutputRange output, ASTInformation* astInformation, Config* config)
+        OutputRange output, ASTInformation* astInformation, Config* config)
     {
         this.rawSource = rawSource;
         this.tokens = tokens;
@@ -150,8 +150,8 @@ private:
         {
             formatComment();
         }
-        else if (isStringLiteral(current.type) || isNumberLiteral(current.type)
-                || currentIs(tok!"characterLiteral"))
+        else if (isStringLiteral(current.type)
+                || isNumberLiteral(current.type) || currentIs(tok!"characterLiteral"))
         {
             writeToken();
             if (index < tokens.length)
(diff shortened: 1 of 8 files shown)


src/dfmt/formatter.d
```diff
@@ -585,7 +591,8 @@ private:
             if (peekBackIs(tok!")"))
                 write(" ");
             auto e = expressionEndIndex(index);
-            immutable bool hasComment = tokens[index .. e].canFind!((a, b) => a.type == b)(tok!"comment");
+            immutable bool hasComment = tokens[index .. e].canFind!((a, b) => a.type == b)(
```
qznc: Is this a sensible point to insert a line break?

src/dfmt/formatter.d
```diff
@@ -471,15 +474,18 @@ private:
         import dfmt.editorconfig : OptionalBoolean;
 
         immutable bool isCase = astInformation.caseEndLocations.canFindIndex(current.index);
-        immutable bool isAttribute = astInformation.attributeDeclarationLines.canFindIndex(current.line);
+        immutable bool isAttribute = astInformation.attributeDeclarationLines.canFindIndex(
```
qznc: Maybe a line break before ".canFindIndex" would be better?

src/dfmt/formatter.d
```diff
@@ -1119,7 +1143,8 @@ private:
                 }
                 immutable l = indents.indentToMostRecent(tok!"switch");
                 if (l != -1)
-                    indentLevel = config.dfmt_align_switch_statements == OptionalBoolean.t ? l : indents.indentLevel;
+                    indentLevel = config.dfmt_align_switch_statements == OptionalBoolean.t ? l
```
qznc: A line break before OptionalBoolean would maybe be better?

src/dfmt/formatter.d
```diff
@@ -1172,8 +1197,8 @@ private:
             }
             else
             {
-                while (indents.topIsTemp() && (peekBackIsOneOf(true, tok!"}", tok!";")
-                        && indents.top != tok!";"))
+                while (indents.topIsTemp() && (peekBackIsOneOf(true, tok!"}",
```
qznc: Might be a bug?

src/dfmt/tokens.d
```diff
@@ -197,20 +196,21 @@ private string generateFixedLengthCases()
         "break", "byte", "case", "cast", "catch", "cdouble", "cent", "cfloat",
         "char", "class", "const", "continue", "creal", "dchar", "debug",
         "default", "delegate", "delete", "deprecated", "do", "double", "else",
```
qznc: Should we use //dfmt off here?

src/dfmt/formatter.d
Hackerpilot: The algorithm is designed to prefer breaking lines at `(` over `.`. I don't think this is a bug.

src/dfmt/tokens.d
Hackerpilot: I don't see how this new formatting is much worse than the old one.

src/dfmt/formatter.d
qznc: Split off into #181  I have a commit, which splits at "=", but it breaks other tests at the moment.

Hackerpilot:
I'm going to close this because:
* It got very out-of-date
* I have run dfmt on itself recently

PR #184, opened 2015-09-08
allow to override dmd

MartinNowak:


Hackerpilot:
Looks like the CI failure is a result of the changes to linking and code generation in 2.068.

Hackerpilot:
Actually this broke between 2.068 and 2.068.1.

Hackerpilot:
I was able to fix it by specifying the path to std.format, but that seems like a hack:

```makefile
SRC := $(shell find src -name "*.d") \
	$(shell find libdparse/src -name "*.d") \
	$(shell find libdparse/experimental_allocator/src -name "*.d") \
	/home/brian/.dvm/compilers/dmd-2.068.1/src/phobos/std/format.d
```

I shouldn't have to specify that Phobos files get compiled, should I?

MartinNowak:
Sorry for not letting your know, we already fixed the issue.
https://trello.com/c/ARtw7Xwl/81-reg2-068-1-linker-error-with-speculative-instantiation-and-inline
Looks like we have to release a 2.068.2 soon.

MartinNowak:
What about this PR?

Hackerpilot:
I ended up making a similar changes, but using the `DC` environment variable because that's what Travis uses.

Hackerpilot:
See https://github.com/Hackerpilot/dfmt/commit/480e6fe973e1c8315259892400d3d77cc82901cb

PR #202, opened 2016-01-14
Add --assume_filename option

JohanEngelen:
1. Add --assume_filename option to control the search path for the style config file when reading from stdin; needed for editor integration.
2. Add Sublime editor integration python script.

Notes:
1. The assume_filename option is needed because the current working directory (cwd) will generally not be the path of the file in the editor. So when the editor passes the file's contents through stdin, without this option, the correct .editorconfig file for the file in the editor would not be found.
2. This Sublime python script is very similar to the clang-format-sublime.py script from the Clang project. Scripts for other editors can be found in clang's repository in ...\tools\clang-format\ . Note that the scripts will have to be simplified because dfmt currently only supports reformatting the whole file.

### DIFF
--- a/editor_integration/dfmt-sublime.py
+++ b/editor_integration/dfmt-sublime.py
@@ -1,0 +1,50 @@
+# This file is a minimal dfmt sublime-integration. To install:
+# - Change 'binary' if dfmt is not on the path (see below).
+# - Put this file into your sublime Packages directory, e.g. on Linux:
+#     ~/.config/sublime-text-2/Packages/User/dfmt-sublime.py
+# - Add a key binding (only for D source files):
+#     { "keys": ["ctrl+shift+c"], "command": "dfmt",
+#       "context": [{ "key": "selector", "operator": "equal", "operand": "source.d" }]
+#     }, 
+#
+# With this integration you can press the bound key and dfmt will
+# format the whole file.
+#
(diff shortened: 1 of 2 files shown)


Hackerpilot:
I used to have editor integration scripts in the DCD repository. It ended up being a bad idea. DCD integration is now handled by separate projects. I want to do the same thing with dfmt. You may want to try adding dfmt support to https://github.com/yazd/DKit.

Hackerpilot:
Wouldn't it be simpler to specify a path to a .editorconfig file?

JohanEngelen:
Implementation-wise, I think it is simpler to specify a file (like was already done with "dummy.d"). For the editor/user, it is nice to specify a file because that is what the editor is passing through stdin. This would then do the exact same thing as if the file was passed through the commandline (instead of stdin). In theory this would allow different editorconfig files for different files (or different file types) in the same directory. When specifying a path+stdin instead, you would not be able to get the same behavior as passing a file.

JohanEngelen:
The editor integration script is meant to be a single script, with no installation requirements and with only dfmt functionality. Perhaps more of an example file for larger editor packages. I like how such scripts are part of the clang-format repository, and not in another separate repo (there are of course many C++ sublime packages out there). This means it is easy for the developers of dfmt/clang-format to test and amend commandline options such as assume_filename.

PR #279, opened 2017-05-17
Add installing with DUB instructions

wilzbach:
Unfortunately currently only the `~master` version builds with a recent DMD installation.


skl131313:
@wilzbach 
Could add another tag, I guess it would be `0.5.0`? Is that all is needed to add a new dub release?

wilzbach:
> Could add another tag, I guess it would be 0.5.0? Is that all is needed to add a new dub release?

Yep, ideally after this PR

https://github.com/dlang-community/dfmt/pull/281

PR #281, opened 2017-05-22
Set DUB config type to `autodetect`

wilzbach:
This simple change and a new preview release tag afterwards is needed to integrate `dfmt` into the DLang Tour (https://github.com/stonemaster/dlang-tour/pull/516) (as a "Format" button).

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -1,7 +1,7 @@
 {
     "name": "dfmt",
     "description": "Dfmt is a formatter for D source code",
-    "targetType": "executable",
+    "targetType": "autodetect",
     "license": "BSL-1.0",
     "dependencies": {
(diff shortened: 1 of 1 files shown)


ZombineDev:
@wilzbach I've just push the 0.5.0 tag

wilzbach:
Thanks a lot for the quick review & merge, guys. Very appreciated!

BBasile:
why ? dfmt is well an executable after all.

wilzbach:
> why ? dfmt is well an executable after all.

And it stays to be, but "autodetect" is a DUB feature which will change the build type to "library" if it's included as dependency to another project as the DLang Tour where `import dfmt.formatter` is called.

BBasile:
Thanks for the clarification.

PR #282, opened 2017-06-10
Enable testing on Windows with AppVeyor

wilzbach:
With the `build.bat`


wilzbach:
This is green now :tada:  - CC @BBasile @ZombineDev

PR #284, opened 2017-07-02
Remove deprecated stdc import

wilzbach:
See also: https://github.com/dlang/phobos/pull/5532

I am doing this blindly (no Windows), but AppVeyor should be able to green-light this.

### DIFF
--- a/tests/issue0114.d
+++ b/tests/issue0114.d
@@ -1,5 +1,5 @@
 private
 {
 	import std.process;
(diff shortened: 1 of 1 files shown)


Hackerpilot:
This file is never built. It's a test input file for the formatter. The builds are failing because you didn't change the expected result files.

BBasile:
Ok so i close, i meant the same thing as HP.

PR #293, opened 2017-10-16
Build debian package using Makd

stefan-koch-sociomantic:
* makd ()...v1.10.1+4(602ca95) (1 commits)
  > Add build flags and flavour to the Version

### DIFF
--- a/makefile
+++ b/makefile
@@ -38,3 +38,8 @@ bin/dfmt: views/VERSION $(SRC)
 debug: views/VERSION $(SRC)
 	$(DC) $(DMD_DEBUG_FLAGS) $(filter %.d,$^) -ofbin/dfmt
 
+pkg: dmd
+	$(MAKE) -f makd/Makd.mak pkg
+
+clean:
+	$(RM) bin/dfmt
--- a/pkg/dfmt.pkg
+++ b/pkg/dfmt.pkg
@@ -1,0 +1,31 @@
+# dfmt packaging configuration
+# ============================
+
+dfmt_bindir='bin'
+
+#-------------------------------------------------------------------------------
+#    Define the keyword arguments to pass to fpm (via the OPTS dict)
+#-------------------------------------------------------------------------------
+
+OPTS.update(
+    name = "dfmt",
+    url = 'https://github.com/dlang-community/dfmt',
+    maintainer = 'Stefan Koch <stefan.koch@sociomantic.com>',
+    vendor = 'Sociomantic Labs GmbH',
+    provides = "dfmt",
+    description = '''\
+D source code formatter
+''',
+    depends = FUN.autodeps('dfmt', path=dfmt_bindir),
+)
+
+#-------------------------------------------------------------------------------
+#    Define the positional arguments to pass to fpm (via the ARGS list)
+#-------------------------------------------------------------------------------
+
+ARGS.extend(FUN.mapfiles(dfmt_bindir, '/usr/bin', 'dfmt'))
+ARGS.extend([
+ 'bash-completion/completion/dfmt=/usr/share/bash-completion/completion/dfmt'
+])
+
+# vim: set ft=python tw=80 :
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "libdparse"]
 	path = libdparse
 	url = https://github.com/dlang-community/libdparse.git
+[submodule "makd"]
+	path = makd
+	url = https://github.com/sociomantic-tsunami/makd


makefile
```diff
@@ -1,4 +1,6 @@
-SRC := $(shell find src -name "*.d") \
+include makd/Makd.mak
```
ZombineDev: I think this should be conditioned on the existence of the `makd` sub-folder, as we don't want to make `makd` a hard dependency. Only users that want/need to use it should explicitly get it.

stefan-koch-sociomantic: good point.

leandro-lucarella-sociomantic: Even better, just use `makd` for packaging by doing something like:
```make
.PHONY: pkg
pkg:
        $(MAKE) -f makd/Makd.mak pkg
```

This way makd will only be used if using the `pkg` target, and will be completely unintrusive for the build process.

ZombineDev: @leandro-lucarella-sociomantic What do you think about obtaining makd on demand? I'm thinking something along the lines of:
```Makefile
pkg: makd
        $(MAKE) -f makd/Makd.mak pkg

makd:
    git clone https://github.com/sociomantic-tsunami/makd
```

ZombineDev: That way the adding makd  should be completely non-intrusive.

makefile
```diff
@@ -15,21 +15,28 @@ GDC ?= gdc
...
-bin/dfmt-test: $(SRC)
+bin/dfmt-test: $(SRCS)
 	$(DC) $(DMD_TEST_FLAGS) $^ -of$@
 
-bin/dfmt: $(SRC)
+bin/dfmt: $(SRCS)
 	$(DC) $(DMD_FLAGS) $^ -of$@
 
-debug: $(SRC)
+debug: $(SRCS)
```
leandro-lucarella-sociomantic: You don't really need to change all these `$(SRC)` to `$(SRCS)` now that makd is not directly included.

makefile
```diff
@@ -15,21 +15,28 @@ GDC ?= gdc
...
 	$(DC) $(DMD_DEBUG_FLAGS) $^ -ofbin/dfmt
+
+pkg: dmd
+	$(MAKE) -f makd/Makd.mak pkg
+
+clean:
+	@rm -rf bin/dfmt
```
leandro-lucarella-sociomantic: It is recommended to use `$(RM)` built in variable (which includes `-f`) and why `-r`? Is it a directory?

So you probably want to use `$(RM) bin/dfmt` instead (I wouldn't use the `@` either).

pkg/defaults.py
```diff
@@ -0,0 +1,18 @@
...
+    vendor = 'Sociomantic Labs GmbH',
+    provides = "dfmt",
+)
+
+#-------------------------------------------------------------------------------
+#    List of fpm positional arguments that are common to all packages
+#-------------------------------------------------------------------------------
+
+#ARGS = [ ]
+
```
leandro-lucarella-sociomantic: This separation is not useful if you only have one package. I would do everything in `dfmt.pkg`.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic do you want me to squash more ?

ZombineDev:
@leandro-lucarella-sociomantic I think you missed my question: What do you think about obtaining makd on demand? I'm thinking something along the lines of:

```Makefile
makd:
    git clone https://github.com/sociomantic-tsunami/makd

pkg: dmd makd
	$(MAKE) -f makd/Makd.mak pkg

clean:
	$(RM) bin/dfmt
```

makefile
```diff
@@ -33,3 +33,10 @@ bin/dfmt: $(SRC)
 
 debug: $(SRC)
 	$(DC) $(DMD_DEBUG_FLAGS) $^ -ofbin/dfmt
+
+pkg: dmd
+	$(MAKE) -f makd/Makd.mak pkg
+
+clean:
+	$(RM) bin/dfmt
```
wilzbach: You could remove the object files and test binary too...

leandro-lucarella-sociomantic:
> @leandro-lucarella-sociomantic I think you missed my question: What do you think about obtaining makd on demand? I'm thinking something along the lines of:

Yeah, that can be done too. Also the rules are incorrect (you can't use directories as dependencies). But we would need to specify a tag when cloning, and the deal with the re-cloning / update when version changes. We basically have to implement `git submodules` ourselves to do it proper, so I don't really see the point :)

makefile
```diff
@@ -33,3 +33,10 @@ bin/dfmt: $(SRC)
 
 debug: $(SRC)
 	$(DC) $(DMD_DEBUG_FLAGS) $^ -ofbin/dfmt
+
+pkg: dmd
```
leandro-lucarella-sociomantic: I would use `bin/dfmt` as dependency instead, that `dmd` target looks super weird, I would expect for `dmd` to be built for a target named like that :D. And also this is what you really want, if for some reason `make dmd` needs to do something else, you don't really care about that.

makefile
leandro-lucarella-sociomantic: Since this project is not using dub for building, I think at some point it would be a good idea to migrate it to Makd, it will take care of all those details :)

stefan-koch-sociomantic: This project DOES dub for building and we actually need to maintain the dub-path as well.
for `dub fetch dfmt && dub run dfmt` to still work

pkg/dfmt.pkg
```diff
@@ -0,0 +1,28 @@
+# waveclean packaging configuration
```
nemanja-boric-sociomantic: Waveclean ;-)

pkg/dfmt.pkg
```diff
@@ -0,0 +1,28 @@
...
+
+dfmt_bindir='bin'
+
+#-------------------------------------------------------------------------------
+#    Define the keyword arguments to pass to fpm (via the OPTS dict)
+#-------------------------------------------------------------------------------
+
+OPTS.update(
+    name = "dfmt",
+    url = 'https://github.com/dlamg-community/dfmt',
```
nemanja-boric-sociomantic: dlamg-community typo.

stefan-koch-sociomantic:
@nemanja-boric-sociomantic addressed.

pkg/dfmt.pkg
```diff
@@ -0,0 +1,28 @@
...
+D source code formatter
+''',
+    depends = FUN.autodeps('dfmt', path=dfmt_bindir),
+)
+
+#-------------------------------------------------------------------------------
+#    Define the positional arguments to pass to fpm (via the ARGS list)
+#-------------------------------------------------------------------------------
+
+ARGS.extend(FUN.mapfiles(dfmt_bindir, '/usr/bin', 'dfmt'))
```
leandro-lucarella-sociomantic: Maybe add the README too?

stefan-koch-sociomantic: I don't need the readme in the target file-system.
but the bash-completion

leandro-lucarella-sociomantic: Is nice to provide a README for users that install the package. Is the standard practice for debian packages, they usually go to `/usr/share/doc/<pkg>/`.

leandro-lucarella-sociomantic:
Also, please squash all the commits together, is really only one conceptual change.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic squashed

leandro-lucarella-sociomantic:
:tada: now we should use beaver to upload the packages to bintray on tags. For example:

https://github.com/sociomantic-tsunami/tangort/blob/5e5293a6b6a49cd73a31119f90dee92580760e02/.travis.yml#L25-L36

PR #294, opened 2017-10-18
Generate version-string from git-describe

stefan-koch-sociomantic:
- adds make target for `views/VERSION` which uses git describe to get the version number
- string imports `views/VERSION` if available and uses it as version-string
- `-debug` is added if in debug

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,8 +1,8 @@
 SRC := $(shell find src -name "*.d") \
 	$(shell find libdparse/src -name "*.d")
 INCLUDE_PATHS := -Ilibdparse/src -Isrc
-DMD_COMMON_FLAGS := -dip25 -w $(INCLUDE_PATHS)
-DMD_DEBUG_FLAGS := -g $(DMD_COMMON_FLAGS)
+DMD_COMMON_FLAGS := -dip25 -w $(INCLUDE_PATHS) -Jviews
+DMD_DEBUG_FLAGS := -debug -g $(DMD_COMMON_FLAGS)
 DMD_FLAGS := -O -inline $(DMD_COMMON_FLAGS)
 DMD_TEST_FLAGS := -unittest -g $(DMD_COMMON_FLAGS)
 LDC_FLAGS := -g -w -oq $(INCLUDE_PATHS)
@@ -14,6 +14,10 @@ GDC ?= gdc
 .PHONY: dmd ldc gdc test
 
 dmd: bin/dfmt
+
+views/VERSION : .git/refs/tags .git/HEAD
+	mkdir -p $(dir $@)
+	git describe --tags > $@
 
 ldc: $(SRC)
 	$(LDC) $(LDC_FLAGS) $^ -ofbin/dfmt
@@ -28,8 +32,9 @@ test: debug
 bin/dfmt-test: $(SRC)
 	$(DC) $(DMD_TEST_FLAGS) $^ -of$@
 
-bin/dfmt: $(SRC)
-	$(DC) $(DMD_FLAGS) $^ -of$@
+bin/dfmt: views/VERSION $(SRC)
+	$(DC) $(DMD_FLAGS) $(filter %.d,$^) -of$@
 
-debug: $(SRC)
-	$(DC) $(DMD_DEBUG_FLAGS) $^ -ofbin/dfmt
+debug: views/VERSION $(SRC)
+	$(DC) $(DMD_DEBUG_FLAGS) $(filter %.d,$^) -ofbin/dfmt
+
--- a/src/dfmt/main.d
+++ b/src/dfmt/main.d
@@ -4,8 +4,61 @@
 //          http://www.boost.org/LICENSE_1_0.txt)
 
 module dfmt.main;
+static immutable VERSION = () {
+    debug
+    {
+        enum DEBUG_SUFFIX = "-debug";
+    }
+    else
+    {
+        enum DEBUG_SUFFIX = "";
+    }
 
-private enum VERSION = "0.5.0";
+    static if (is(typeof(import("VERSION"))))
+    {
+        // takes the `git describe --tags` output and removes the leading
+        // 'v' as well as any kind of newline
+	// if the tag is considered malformed it gets used verbatim
+
+        enum gitDescribeOutput = import("VERSION");
+
+	string result;
+
+        if (gitDescribeOutput[0] == 'v')
+            result = gitDescribeOutput[1 .. $];
+        else
+            result = null;
+
+	uint minusCount;
+
+        foreach (i, c; result)
+        {
+            if (c == '\n' || c == '\r')
+            {
+                result = result[0 .. i];
+                break;
+            }
+
+	    if (c == '-')
+	    {
+                ++minusCount;
+            }
+        }
+
+        if (minusCount > 1)
+		result = null;
+
+        return result ? result ~ DEBUG_SUFFIX
+                      : gitDescribeOutput ~ DEBUG_SUFFIX;
+
+    }
+    else
+    {
+        return "unknown"  ~ DEBUG_SUFFIX ~ "-version";
+    }
+
+} ();
+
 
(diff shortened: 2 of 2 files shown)


views/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!.gitignore
```
leandro-lucarella-sociomantic: I don't think you need to specify `.gitignore` because AFAIK `*` doesn't match files starting with `.`.

In any case, it seems more sensible IMHO to just put `views/` in the top-level `.gitignore` instead.

BTW, what is this folder for? Why it exists and why should it be ignored, some clarification in the commit message would be helpful!

makefile
```diff
@@ -1,7 +1,7 @@
 SRC := $(shell find src -name "*.d") \
 	$(shell find libdparse/src -name "*.d")
 INCLUDE_PATHS := -Ilibdparse/src -Isrc
-DMD_COMMON_FLAGS := -dip25 -w $(INCLUDE_PATHS)
+DMD_COMMON_FLAGS := -dip25 -w $(INCLUDE_PATHS) -Jviews
```
leandro-lucarella-sociomantic: Oh, I see now the "views" is for string inclusion. Why is it called "views" though? It looks like a funny name. But please squash the change adding the directory to the `.gitignore` file with this one.

makefile
```diff
@@ -15,6 +15,9 @@ GDC ?= gdc
 
 dmd: bin/dfmt
 
+views/VERSION : $(SRC)
+	git describe --tags | cut -c2- | cut -f1-2 -d- > views/VERSION
```
leandro-lucarella-sociomantic: This needs a comment, what are you cutting and why? Also, use `> $@` instead of `> views/VERSION`, so if for some reason the file is renamed you don't need to update the script too.

leandro-lucarella-sociomantic: Also, this probably needs to get built unconditionally. If you change a README file and commit for example, the git hash /tag you are building will change, but the code will say you built another version. While the binary would be the same, it seems misleading.

You can unconditionally generate this file every time `make` runs via this trick:

```make
__dummy_version := $(shell git describe --tags | cut -c2- | cut -f1-2 -d- > views/VERSION)
```

makefile
```diff
@@ -28,8 +31,9 @@ test: debug
 bin/dfmt-test: $(SRC)
 	$(DC) $(DMD_TEST_FLAGS) $^ -of$@
 
-bin/dfmt: $(SRC)
-	$(DC) $(DMD_FLAGS) $^ -of$@
+bin/dfmt: views/VERSION
+	$(DC) $(DMD_FLAGS) $(SRC) -of$@
```
leandro-lucarella-sociomantic: Better filter files: `$(DC) $(DMD_FLAGS) $(filter %.d,$^) -of$@`

makefile
```diff
@@ -28,8 +31,9 @@ test: debug
...
+
+debug: views/VERSION
+	$(DC) $(DMD_DEBUG_FLAGS) $(SRC) -ofbin/dfmt
```
leandro-lucarella-sociomantic: Same.

makefile
```diff
@@ -2,7 +2,7 @@ SRC := $(shell find src -name "*.d") \
 	$(shell find libdparse/src -name "*.d")
 INCLUDE_PATHS := -Ilibdparse/src -Isrc
 DMD_COMMON_FLAGS := -dip25 -w $(INCLUDE_PATHS) -Jviews
-DMD_DEBUG_FLAGS := -g $(DMD_COMMON_FLAGS)
+DMD_DEBUG_FLAGS := -debug -g $(DMD_COMMON_FLAGS)
```
leandro-lucarella-sociomantic: Please do this in a separate commit.

src/dfmt/main.d
```diff
@@ -5,7 +5,25 @@
...
+
+static if (is(typeof(import("VERSION"))))
+{
+    import std.string : chomp;
+    private enum VERSION = chomp(import("VERSION")) ~ DEBUG_SUFFIX;
+}
+else
+{
+    private enum VERSION = "0.6.x" ~ DEBUG_SUFFIX;
+}
```
leandro-lucarella-sociomantic: I think this should fail if `VERSION` can't be found. This will get outdated extremely soon otherwise, what's the point on keeping a version like this?

views/.gitignore
stefan-koch-sociomantic: I need the .gitigonre because git ignores empty directories.
I cannot add views if it has no files.
And views need to exist for `-J` to work

leandro-lucarella-sociomantic:
```make
__dummy_version := $(shell mkdir -p views && git describe --tags | cut -c2- | cut -f1-2 -d- > views/VERSION)
```

stefan-koch-sociomantic: I don't like this :-(

makefile
stefan-koch-sociomantic: `views` is the default folder dub uses for string inclusion

views/.gitignore
leandro-lucarella-sociomantic: Why?

makefile
leandro-lucarella-sociomantic: OK then. I wonder where does that comes from, but if it's some sort of defacto standard, I'm fine with it.

makefile
stefan-koch-sociomantic: `$^` are the dependencies right ? in this case `views/VERSION`
that would not work

makefile
stefan-koch-sociomantic: It comes from the fact that vibed's `diet-templates` are in there.
which are the view in M-V-C.

leandro-lucarella-sociomantic: Typical "historical reasons" :man_facepalming:

makefile
leandro-lucarella-sociomantic: `$^` are all the dependencies, yes, but `$(filter %.d,$^)` are only the dependencies matching `*.d` so to say, which is exactly what you want to pass to DMD, so it will work, it does for Makd :)

stefan-koch-sociomantic: this expands to
`dmd -O -inline -dip25 -w -Ilibdparse/src -Isrc -Jviews  -ofbin/dfmt`
since views/VERSION is the only dependency

views/.gitignore
stefan-koch-sociomantic: it looks hacky :)

src/dfmt/main.d
stefan-koch-sociomantic: make is not the only build-system used;
The dub path has to work as well; I cannot figure out howto make writing of the version file work on windows.

stefan-koch-sociomantic:
The travis error comes from dmd-nightly not being downloaded.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic please have another look

makefile
leandro-lucarella-sociomantic: NOT addressed yet.

src/dfmt/main.d
```diff
@@ -4,8 +4,61 @@
...
+        if (minusCount > 1)
+		result = null;
+
+        return result ? result ~ DEBUG_SUFFIX
+                      : gitDescribeOutput ~ DEBUG_SUFFIX;
+
+    }
+    else
+    {
+        return "0.6.x" ~ DEBUG_SUFFIX;
```
leandro-lucarella-sociomantic: I don't recall why is this still there. Can't you make `git describe > views/VERSION` work for all platform so no manual version has to be maintained?

leandro-lucarella-sociomantic:
```
$ source "$(CURL_USER_AGENT="$CURL_USER_AGENT" bash install.sh dmd-nightly --activate)"
Warning: No gpg tool found to verify downloads.
curl: (56) Recv failure: Operation timed out
/Users/travis/.travis/job_stages: line 57: : No such file or directory
```

I'll re-run :crossed_fingers:

PR #298, opened 2017-10-26
[UrgentBugifx] Fix space before function parameters

stefan-koch-sociomantic:
This fixes a bad bug where `--space-before-function-parameters=true` would add a `' '` before every identifier followed by a `(`
so `abc(` would become `abc (` regardless of being delcaration or call.

this is an urgent fix @leandro-lucarella-sociomantic @rikkimax

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -281,8 +281,8 @@ private:
         {
             writeToken();
             if (index < tokens.length && (currentIs(tok!"identifier")
-                    || ( ( isBasicType(peekBack().type) || peekBackIs(tok!"identifier") ) &&
-                         currentIs(tok!("(")) && config.dfmt_space_before_function_parameters)
+                    || ( index < 1 && ( isBasicType(peekBack(2).type) || peekBack2Is(tok!"identifier") ) &&
+                         currentIs(tok!("(")) && config.dfmt_space_before_function_parameters )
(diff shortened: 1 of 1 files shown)


src/dfmt/formatter.d
```diff
@@ -281,8 +281,8 @@ private:
         {
             writeToken();
             if (index < tokens.length && (currentIs(tok!"identifier")
-                    || ( ( isBasicType(peekBack().type) || peekBackIs(tok!"identifier") ) &&
```
Hackerpilot: I'm guessing that this sub-expression would always return true because we just wrote an identifier, and therefore `peekBackIs(tok!"identifier")` is always true. Is that the cause of the bug?

stefan-koch-sociomantic: Yes exactly.

Hackerpilot:
Strange that this is causing assertion failures, but only on Mac.

stefan-koch-sociomantic:
since this only fails on mac it might be a codegen bug.
They ... I'd need a mac to check it out.

stefan-koch-sociomantic:
hmm seems like the `assert` was ignored on every platform but mac.
Still I cannot reproduce the failure locally.

Hackerpilot:
Bypassing the CI check because the only thing that's not working is the nightly DMD build, and that failure is not related to this pull request at all.

PR #302, opened 2017-11-20
[WIP] Implement import-sorting

stefan-koch-sociomantic:
This implements the `--sort_imports` switch which will sort imports alphabetically.

However this functionality still has bugs.
For example will it promote imports out of anonymous scopes in the closest named-scope.
usually a function definition.


WebFreak001:
how would this sort

```d
import b;
import a;
void foo() {}
import c;
void bar() {}
```

or

```d
import b;
import a;

import c;
```

I think if there is something between a block of imports it should not mix them together. I would get rather irritated if dfmt destroyed my grouped sorted imports. I just want phobos imports to always be at the top, no matter if there is a package starting with a.

Also you mixed in tabs a few times and there is still a writeln call, you should run dfmt over the code and drop that debug call before merging.

stefan-koch-sociomantic:
@WebFreak001 
Currently it'll sort 
```
import a;
import b;
import c;
import d;
```

Thanks for making me aware of tabs the writeln

stefan-koch-sociomantic:
also it'll currently drop the private, public and the static in form the import

WebFreak001:
I think it's bad if it rips them out of context. I implemented a sort imports command once which takes into account blocks. I am not sure if it keeps public/private/static, but you can look at the source here: https://github.com/Pure-D/workspace-d/blob/master/source/workspaced/com/importer.d#L85-L115 and here: https://github.com/Pure-D/workspace-d/blob/master/source/workspaced/com/importer.d#L316-L383

WebFreak001:
this definitely needs some unittests

wilzbach:
Have a look at https://github.com/wilzbach/dscanner-fix/blob/master/README.md for an already working solution - it's a lot easier to fix code based on the warnings from DScanner directly.
(Though this project is a weekend project and I stupidly parsed the text output from DScanner)

stefan-koch-sociomantic:
@WebFreak001 indeed it does. currently it'll still remove some imports. :)

leandro-lucarella-sociomantic:
Changed base branch to v0.6.x (as agreed on https://github.com/dlang-community/dfmt/issues/320).

leandro-lucarella-sociomantic:
Changed base again, now to v0.x.x (sorry, v0.6.x is only for bug fixes, my bad).

BBasile:
sorting is not formating

PR #304, opened 2017-11-25
Add support for `static foreach`

### LINKED ISSUE #303: Wrong indentation for `static foreach`
```d
static foreach (thing; things)
        {
        doStuff();
    }
static foreach_reverse (thing; things)
        {
        doStuff();
    }
static foreach (thing; things)
        doStuff();

static foreach_reverse (thing; things)
        doStuff();
```

sould be

```d
static foreach (thing; things)
{
    doStuff();
}
static foreach_reverse (thing; things)
{
    doStuff();
}
static foreach (thing; things)
    doStuff();

static foreach_reverse (thing; things)
    doStuff();
```

kotet:
Fix #303

### DIFF
--- a/src/dfmt/indentation.d
+++ b/src/dfmt/indentation.d
@@ -197,7 +197,8 @@ private:
                 immutable currentIsNonWrapTemp = !isWrapIndent(arr[i])
                     && isTempIndent(arr[i]) && arr[i] != tok!")" && arr[i] != tok!"!";
                 if (arr[i] == tok!"static" && (arr[i + 1] == tok!"if"
-                        || arr[i + 1] == tok!"else") && (i + 2 >= index || arr[i + 2] != tok!"{"))
+                        || arr[i + 1] == tok!"else" || arr[i + 1] == tok!"foreach"
+                        || arr[i + 1] == tok!"foreach_reverse") && (i + 2 >= index || arr[i + 2] != tok!"{"))
                 {
                     parenCount = pc;
                     continue;
--- a/src/dfmt/indentation.d
+++ b/src/dfmt/indentation.d
@@ -176,6 +176,7 @@ private:
 
     int indentSize(const size_t k = size_t.max) const pure nothrow @safe @nogc
     {
+        import std.algorithm : among;
         if (index == 0 || k == 0)
             return 0;
         immutable size_t j = k == size_t.max ? index : k;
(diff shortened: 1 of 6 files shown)


src/dfmt/indentation.d
```diff
@@ -197,7 +197,8 @@ private:
                 immutable currentIsNonWrapTemp = !isWrapIndent(arr[i])
                     && isTempIndent(arr[i]) && arr[i] != tok!")" && arr[i] != tok!"!";
                 if (arr[i] == tok!"static" && (arr[i + 1] == tok!"if"
-                        || arr[i + 1] == tok!"else") && (i + 2 >= index || arr[i + 2] != tok!"{"))
+                        || arr[i + 1] == tok!"else" || arr[i + 1] == tok!"foreach"
+                        || arr[i + 1] == tok!"foreach_reverse") && (i + 2 >= index || arr[i + 2] != tok!"{"))
```
WebFreak001: I would `import std.algorithm : among` and make this
```d
arr[i + 1].among!(tok!"if", tok!"else", tok!"foreach", tok!"foreach_reverse")
```
because it has a lot of cases now

tests/issue0303.d
```diff
@@ -0,0 +1,4 @@
+static foreach (thing; things){doStuff();}
+static foreach_reverse (thing; things){doStuff();}
+static foreach (thing; things) doStuff();
+static foreach_reverse (thing; things) doStuff();
```
WebFreak001: This will have grammar errors after upgrading libdparse (which isn't a big problem because the unittest works, but it will eventually make the test output unreadable). Imagine the static foreach would be a static if, you can't just have function calls at a global level. Please put them in some function like `void main() {}` or change the body of the foreachs to declarations

kotet:
I confirmed that v0.7.2-alpha.3 works, but it's still an alpha version. Is that OK?

WebFreak001:
well if nobody else replies to this soon I will just update it to that. I don't think it really matters that much anyway as long as you use a tagged release, but I am not really active in the libdparse development so I don't really know if there are any changes in API or features someone wanted to address in this repo etc.

wilzbach:
Libdparse hasn't received any breaking changes. It only gets updated with the grammar being changed from time to time and sometimes to fix new compiler warnings.

WebFreak001:
@wilzbach is it ok if I push a libdparse update directly to master then?

wilzbach:
Yup. We often do this for DScanner

WebFreak001:
@wilzbach hm I don't have permissions to push to master, do you want to do it or should I open a PR for that?

wilzbach:
No one has the permission to push to master. Only passing PRs can be merged...

kotet:
I updated libdparse in this PR.

kotet:
It seems that Travis CI failed to install the compiler.

WebFreak001:
you didn't update the dependency in the dub.json

kotet:
I rebased the commit.

Issue #301, opened 2017-11-14
Blank lines inserted before private imports


Barugon:
Before dfmt:
```D
private import std.file;
private import std.path;
import std.stdio;
private import std.string;
```

After dfmt:
```D
private import std.file;

private import std.path;
import std.stdio;

private import std.string;
```

stefan-koch-sociomantic:
@Barugon what would be your expectation for imports to be formatted ?

Barugon:
If they're not private then a blank line is not inserted.

Before dfmt:
```D
import std.file;
import std.path;
import std.stdio;
import std.string;
```

After dfmt:
```D
import std.file;
import std.path;
import std.stdio;
import std.string;
```

I would expect the behavior to be the same for private imports. If I make them all private then this is what I get after using dfmt:
```D
private import std.file;

private import std.path;

private import std.stdio;

private import std.string;
```

stefan-koch-sociomantic:
I have now run into this as well.
I am going to fix this; however it's not clear to me why this actually happens ;)

stefan-koch-sociomantic:
This is closed via #305

PR #305, opened 2017-11-28
Fix issue #301

stefan-koch-sociomantic:
By adding private and static as valid import prefix


stefan-koch-sociomantic:
I am self merging this because of triviality :)

wilzbach:
Please only merge stuff after one day, s.t. everyone had a chance to look at it - even if it's trivial in your eyes. For example, your PR came without tests ...

stefan-koch-sociomantic:
> Please only merge stuff after one day, s.t. everyone had a chance to look at it - even if it's trivial in your eyes. For example, your PR came without tests ...

I plead guilty.
It fixed a bug which tests would have caught if they had been there:)

PR #307, opened 2017-12-01
Sbfp fix

stefan-koch-sociomantic:
This fixes the --space-before-function-paramters to also work on functions returning slices and to also put the space on constructors which are functions as well.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1,3 +1,4 @@
+
 //          Copyright Brian Schott 2015.
 // Distributed under the Boost Software License, Version 1.0.
 //    (See accompanying file LICENSE.txt or copy at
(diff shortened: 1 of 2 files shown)


src/dfmt/formatter.d
```diff
@@ -15,6 +16,9 @@ import dfmt.tokens;
 import dfmt.wrapping;
 import std.array;
 
+extern (C) void breakpoint2017 () {}
```
Hackerpilot: What does this do?

src/dfmt/formatter.d
```diff
@@ -280,12 +296,18 @@ private:
...
+            //dfmt off
             if (index < tokens.length && (currentIs(tok!"identifier")
-                    || ( index > 1 && ( isBasicType(peekBack(2).type) || peekBack2Is(tok!"identifier") ) &&
-                         currentIs(tok!("(")) && config.dfmt_space_before_function_parameters )
+                    || ( index > 1 && ( isBasicType(peekBack(2).type)
+                                        || peekBack2Is(tok!"identifier")
+                                        || peekBack2Is(tok!")")
+                                        || peekBack2Is(tok!"]") )
+                        && currentIs(tok!("("))
+                        && config.dfmt_space_before_function_parameters )
```
Hackerpilot: `config.dfmt_space_before_function_parameters` is the expression that's probably the cheapest to evaluate and also the most likely to be false. It's probably a good idea to put that first in this chain of `&&`s.

stefan-koch-sociomantic:
@Hackerpilot I fixed the messed up commits

PR #309, opened 2017-12-07
Put a space behind static and shared constructors as well

stefan-koch-sociomantic:
This is a change which allows to keep track of all kinds of contructors/destructors which in turn allows --space_before_function_parameters to work.

### DIFF
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -31,6 +31,8 @@ struct ASTInformation
         sort(contractLocations);
         sort(constraintLocations);
         sort(constructorDestructorLocations);
+        sort(staticConstructorDestructorLocations);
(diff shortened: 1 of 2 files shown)


rikkimax:
I'm happy for this to be pulled once CI completes.

Unless anybody else has any objections?

wilzbach:
> I'm happy for this to be pulled once CI completes.

Are you aware that this is why we have the auto-merge label?

stefan-koch-sociomantic:
@rikkimax sorry this is actually not ready yet ... these things need to be treated diffrently.

stefan-koch-sociomantic:
@rikkimax this is ready now.

stefan-koch-sociomantic:
tests fail because of the ldc2 binary not being there on app-veyor

Issue #308, opened 2017-12-07
Integrate dfmt with beaver


stefan-koch-sociomantic:
dfmt should be using beaver for automatic building and distribution of packages with bintray.

stefan-koch-sociomantic:
this is done by #311

PR #311, opened 2017-12-12
Deploy tagged versions from CI

stefan-koch-sociomantic:
This adds automatic deployment (using https://github.com/sociomantic-tsunami/beaver) as soon as new versions are tagged.

This addresses #308

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "makd"]
 	path = makd
 	url = https://github.com/sociomantic-tsunami/makd
+[submodule "beaver"]
+	path = beaver
+	url = https://github.com/sociomantic-tsunami/beaver.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,30 @@ env:
   - BUILD=dub
 
 script: ./.travis.sh
+
+jobs:
+    include:
+        - stage: Build & Upload Package
+          if: tag IS present
+          # Which package to deploy
+          env:
+              - DMD=2.076.*
+              - DIST=xenial
+              - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
+          language: generic
+          sudo: required
+          services:
+              - docker
+          git:
+              submodules: false
+          before_install: git submodule update --init
+          install: beaver dlang install
+          script:
+              - beaver dlang make pkg
+          deploy:
+              provider: script
+              script: beaver bintray upload build/last/pkg/*.deb
+              skip_cleanup: true
+              on:
+                  tags: true # must be a git tag
+                  repo: dlang-community/dfmt # must be a tag on dlang-community
--- a/beaver.Dockerfile
+++ b/beaver.Dockerfile
@@ -1,0 +1,6 @@
+# Copyright sociomantic labs GmbH 2017.
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+FROM sociomantictsunami/dlang:v4


.travis.yml
```diff
@@ -1,18 +1,25 @@
 sudo: false
 language: d
-d:
-  - dmd-nightly
-  - dmd-beta
-  - dmd
-  - ldc-beta
-  - ldc
```
wilzbach: Ehm why?

.travis.yml
```diff
@@ -1,18 +1,25 @@
...
-  - osx
 
 env:
-  - BUILD=
-  - BUILD=dub
+    matrix:
+        - BUILD=
+        - BUILD=dub
+    global:
+        - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
```
wilzbach: Use the matrix `include` feature if you want to run this on a single build. Please don't kill all the other compiler tests.

stefan-koch-sociomantic:
ah right I am going to re-enable the build-matrix :)

.travis.yml
leandro-lucarella-sociomantic: Yeah, this is not needed at all, all you need to do for packaging can be done in the stage that runs only on tags.

.travis.yml
```diff
@@ -1,18 +1,25 @@
...
 
-script: ./.travis.sh
+install: beaver docker build .
+
+script: 
+    - beaver docker run make pkg
```
leandro-lucarella-sociomantic: All these should go straight into a separate stage. Please see https://github.com/sociomantic-tsunami/dmqnode/blob/7b75018e43aeebca4eed0e86e6b2d4b8291124da/.travis.yml#L34-L45 for an example...

Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Copyright sociomantic labs GmbH 2017.
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+FROM sociomantictsunami/dlang:xenial-v3
```
leandro-lucarella-sociomantic: This should also be just `:v3` and `DIST` should be defined as an env variable, all in the separate stage.

docker/build
```diff
@@ -2,17 +2,7 @@
...
-    mv dfmt ./bin
-elif [[ $DC == ldc2 ]]; then
-    git submodule update --init --recursive
-    make ldc -j2
-else
-    git submodule update --init --recursive
-    make pkg -j2
-fi
+git submodule update --init --recursive
+make pkg -j2
```
leandro-lucarella-sociomantic: This doesn't belong to `docker/build`, again this is just to build the docker image, not the project!

.travis.yml
```diff
@@ -19,3 +19,16 @@ env:
...
 script : .travis.sh
+
+jobs:
+    include:
+        - stage: Upload Package
+          # We need to include the exclusion of D2 tags because this "if"
+          # replaces the global one
+          if: tag IS present
+          # Which package to deploy
+          env: DMD=dmd DIST=xenial
```
leandro-lucarella-sociomantic: Use a specific version in here: `DMD=2.077.*` for example, so you don't have to depend on the image's default version.

.travis.yml
```diff
@@ -19,3 +19,16 @@ env:
...
+          # before_install: and install: are inherited and we don't need to
+          # override them.
+           sudo: required
+           script : beaver docker build . && beaver docker run make pkg && beaver bintray upload -N build/last/pkg/*.deb
```
leandro-lucarella-sociomantic: Ideally `beaver docker build` should go to the `install:` part. Some other parts should probably be moved from the general config to the stage too, to avoid messing with the more pure-travis approach of the rest of the `.travis.yml` file, for example the `before_install:` and `git:` parts (not sure if all can be moved there as I never tried it myself). You are missing the `-d` argument to the upload command. Also to build the package you should use `beaver dlang make`, not `run make`.
```yml
jobs:
    include:
        - stage: Build & Upload Package
          if: tag IS present
          # Which package to deploy
          env:
              - DMD=2.077.*
              - DIST=xenial
              - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
          language: generic
          sudo: required
          services:
              - docker
          git:
              submodules: false
          before_install: git submodule update --init
          install: beaver docker build .
          script:
              - beaver docker run make pkg
              - beaver bintray upload -N -d dlang-community/apt build/last/pkg/*.deb
```

wilzbach:
> After this looks good, then you can easily test it by creating a -rc.1 release (or maybe -beta.1 release if you don't feel confident). But don't do that until the PR gets reviewed and approved.

It's also possible enable Travis for ones fork. This has the advantage of being able to push as many tags as wanted there ;-)

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic I pushed few commits accidentally; will clean it up as soon as it works.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic I have squashed the commits.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "makd"]
 	path = makd
 	url = https://github.com/sociomantic-tsunami/makd
+[submodule "beaver"]
+	path = beaver
+	url = https://github.com/sociomantic-tsunami/beaver.git
```
leandro-lucarella-sociomantic: Beaver is now at 0.3.2, I recommend using the latest version.

leandro-lucarella-sociomantic:
I'd also recommend doing some test in your fork, make a tag and probably you need to remove the `on: repo: ...` condition, so the job is executed anyway (I'm not sure how useful it to keep that since the upload will fail anyways if you don't have the right credentials). In the PR tests the deploy stage is never executed.

stefan-koch-sociomantic:
I just switched back to 2.072. since 2.077 was not getting installed. apparently d-apt changed thier naming scheme.

leandro-lucarella-sociomantic:
Use 2.076.* then?

.travis.yml
```diff
@@ -1,4 +1,7 @@
-sudo: false
+sudo: required
+services:
+- docker
```
wilzbach: Remove

.travis.yml
```diff
@@ -1,4 +1,7 @@
-sudo: false
+sudo: required
```
wilzbach: Remove

.gitmodules
leandro-lucarella-sociomantic: There is v0.3.3 already, maybe update before merging? It only fixes a problem when trying to use DMD v2.077+ (to account for a package name change in the d-apt repo).

.travis.yml
```diff
@@ -12,7 +15,37 @@ os:
   - osx
 
 env:
-  - BUILD=
-  - BUILD=dub
+    matrix:
+        - BUILD=
+        - BUILD=dub
+    global:
+        - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
```
leandro-lucarella-sociomantic: This `PATH` is also in the `jobs: include:` I guess one of them should go.

beaver.Dockerfile
```diff
@@ -0,0 +1,6 @@
+# Copyright sociomantic labs GmbH 2017.
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+FROM sociomantictsunami/dlang:v3
```
leandro-lucarella-sociomantic: Completely optional, but there is v4 now.

leandro-lucarella-sociomantic:
> This looks generally OK to me. As long as this sticks to it's separate matrix build job, I think no one bothers much about it. FYI: If it ever breaks and you aren't active anymore, we will probably simply remove the Travis job.

Sounds fair!

leandro-lucarella-sociomantic:
Also, now that v2.076 is used, does it make sense to advance libdparse? I guess it doesn't hurt either.

.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
```
leandro-lucarella-sociomantic: Why is `sudo: false` removed?

.travis.yml
```diff
@@ -12,7 +11,35 @@ os:
   - osx
 
 env:
-  - BUILD=
-  - BUILD=dub
+    matrix:
+        - BUILD=
+        - BUILD=dub
```
leandro-lucarella-sociomantic: Why adding the `matrix:` subitem?

leandro-lucarella-sociomantic:
@wilzbach OSX builds seem to never progress. Do you think they should be kept?

wilzbach:
Rebased as another PR updated libdparse too.

wilzbach:
> Do you think they should be kept?

Ehm Travis seems to be having huge issues with OSX at the moment. For the time being we can use the admin override if all other tests pass, but if the situation continues to be like this I guess we have no other choice.

wilzbach:
I used the admin override, tagged 0.5.2, cancelled the stalled Travis builds in the queue and set the cron to weekly.
@stefan-koch-sociomantic you might want to have a look at:

https://travis-ci.org/dlang-community/dfmt/builds/333380979

leandro-lucarella-sociomantic:
OK, the problem with OSX building being stalled is if they never happen, the deployment stage will never be reached.

wilzbach:
Argh - yeah let's remove them then.
There's much platform-specific code in `dfmt` anyways.

leandro-lucarella-sociomantic:
@stefan-koch-sociomantic can you take care of it?

PR #316, opened 2018-01-25
Make alias visit public in src/dfmt/ast_info.FormatVisitor

RazvanN7:
Making the visiting alias private makes it so that the visit(Module) overload is not visible outside the FormatVisitor class. The code worked up until now due to a bug [1] in the compiler.

[1] dlang/dmd#7766


wilzbach:
The Travis build queue is still stalled for OSX :/

PR #317, opened 2018-01-26
Disable osx

stefan-koch-sociomantic:
as requested in #311

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ d:
 
 os:
   - linux
-#  - osx # disabled since until travis has more mac resources
+#  - osx # disabled until travis has more mac resources
 
 env:
   - BUILD=
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ d:
 
 os:
   - linux
-  - osx
+#  - osx # disabled since until travis has more mac resources
 
 env:
   - BUILD=


leandro-lucarella-sociomantic:
I wonder if we should disable tests in ldc in appveyor too, ldc is tested by travis anyway and it just makes looking at the PR status useless...

leandro-lucarella-sociomantic:
Or fix it, it seems to be trying to fetch a non existing file: https://github.com/ldc-developers/ldc/releases/download/v1.7.0/ldc2-1.7.0-win64-msvc.zip

But in any case leaving it when it fails consistently seems pretty pointless.

wilzbach:
> Or fix it, it seems to be trying to fetch a non existing file:

Yes it needs this diff - https://github.com/dlang/dub/pull/1332

> I wonder if we should disable tests in ldc in appveyor too, 

I wouldn't mind. AppVeyor's slowness is annoying.
Just dmd x86 + x86_64 should be enough (+optionally ldc).

leandro-lucarella-sociomantic:
Well, if it's reasonably slow (like an hour or so) and it works and it's a simple fix, I don't mind keeping AppVeyor.

leandro-lucarella-sociomantic:
What's more annoying for me is the failure that makes the final github status pretty useless.

PR #322, opened 2018-02-02
Fix #321

### LINKED ISSUE #321: dfmt doesn't work with `#line` primitive
before:
```
void test(){
  #line 100
  int a;
  #line 200 "anotherfile"
  int b;
}
```

after:
```
void test()
{
    #line 100int a;
    #line 200 "anotherfile"int b;
}
```

Hackerpilot:


wilzbach:
AppVeyor problems are due to a  DUB timing out - 2.078.2 will contain a fix for this. Testing dmd-nightly and dmd-beta on AppVeyor is definitely an overkill though: https://github.com/dlang-community/dfmt/pull/323
(I should really get https://github.com/dlang/installer/pull/275 going, s.t. we don't have to update these scripts all over the place, but my Windows foo isn't existent)

PR #331, opened 2018-02-11
Bump dependencies

wilzbach:


BBasile:
This is not 100% necessary. alpha.4 was already taken b/c of libdparse and accepted for dfmt because of the `~>` operator.

wilzbach:
> This is not 100% necessary. 

It is for the git submodules :/

BBasile:
There's a bat in there too. i Will push.

PR #335, opened 2018-02-21
Change to more effcient optionsToString template

stefan-koch-sociomantic:
This reduces compiletime quite dramatically by (0.3 seconds).

### DIFF
--- a/src/dfmt/main.d
+++ b/src/dfmt/main.d
@@ -289,20 +289,19 @@ private version (Windows)
     }
 }
 
-private string optionsToString(E)() if (is(E == enum))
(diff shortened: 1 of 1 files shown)


wilzbach:
FWIW dlang-bot can't merge PRs due to @leandro-lucarella-sociomantic new branch settings:

image

Which is a __good__ thing because it prevents SELF PULLS which we agreed you shouldn't DO.

leandro-lucarella-sociomantic:
I had to roll those restrictions back because it conflicts with the release script (we need to push the cleared notes for example, and it doesn't make sense to go through a PR and approval process for such a trivial and automated change).

PR #336, opened 2018-02-21
Fix bad formatting in main.d

stefan-koch-sociomantic:
as per title


BBasile:
Maybe the boat had reach the API limit.

PR #337, opened 2018-02-21
fix #236 - Allow constraints to be indented by a single tab

### LINKED ISSUE #236: Allow constraints to be indeted by a single tab
``` D
void foo(A, Z, E, R, T, Y, U, I)(A a, Z z, E e, R r, T t, Y y, U u, I i)
if (verifySomething!(A, Z, E, R, T, Y, U, I))
{}
```

yields

``` D
void foo(A, Z, E, R, T, Y, U, I)(A a, Z z, E e, R r, T t, Y y, U u, I i)
        if (verifySomething!(A, Z, E, R, T, Y, U, I))
{
}
```

The indentation should made of 4 spaces and not 8 (with default setting so _conditional_newline_indent_).

BBasile:
Constraints were always indented by two TAB (or equivalent in spaces)


BBasile:
Unless this was on purpose since i saw that the tail of long function declarations are also indented by 2.

Hackerpilot:
It was on purpose, mostly because of how the wrapping interacts with the OTBS style:

```
void doStuff(T)(T t)
    if (ThingsAreCool!T) {
    doThings();
}
```
Compare 
```
XXXX XXXXXXXXXXXXXXX
    XX XXXXXXXXXXXXXXXXX X
    XXXXXXXXXXX
X
```
to
```
XXXX XXXXXXXXXXXXXXX
        XX XXXXXXXXXXXXXXXXX X
    XXXXXXXXXXX
X
```

It's easier to see from the extra indent level where the function body actually is.

BBasile:
Ah it's an enhancement. Can i add a new option then ?

BBasile:
Converted to a new option but i don't know how to test this.

UplinkCoder:
This does not look right to me, are you sure this has no unwanted side-effects ?

BBasile:
> are you sure this has no unwanted side-effects ?

No,  the only thing i'm sure of is that it doesn't break the test suite. Note that initially i changed the default behavior (though 2 tabs was a bug) so i had patched all the tests and it worked... All the constraints were 4 spaces indented and split to a second line when reaching the hard right edge limit.

Hackerpilot:
> No, the only thing i'm sure of is that it doesn't break the test suite.

This is how I work on `dfmt`. I'm not sure there's another way.

BBasile:
I've added a system to test features, individually from the main test suite since it's axes around the braces style.

README.md
```diff
@@ -105,6 +106,7 @@ dfmt_space_before_function_parameters | `true`, `false` | `false` | Insert space
 dfmt_selective_import_space | `true`, `false` | `true` | Insert space after the module name and before the `:` for selective imports.
 dfmt_compact_labeled_statements | `true`, `false` | `true` | Place labels on the same line as the labeled `switch`, `for`, `foreach`, or `while` statement.
 dfmt_template_constraint_style | `conditional_newline_indent` `conditional_newline` `always_newline` `always_newline_indent` | `conditional_newline_indent` | Control the formatting of template constraints.
+dfmt_single_template_constraint_indent | `true`, `false` | `false` | Set if the constraints are indented by a single tab instead of two. Has only an effect for certain indentation styles.
```
Hackerpilot: `Set to indent template constraints by a single indent level instead of the default of two.` Also, this should specify which indent styles it affects.

tests/individual/single_constraint_indent.d
```diff
@@ -0,0 +1,73 @@
...
+    if (r == 0)
+    {
+        assert(e.length == 0, "unexpected stderr content: \n" ~ e);
+        const auto formatted_output = "\n" ~ pp.stdout.byLineCopy.array.join("\n");
+        assert(formatted_output == expected, formatted_output);
+    }
+    else assert(false, "abnormal dfmt termination : " ~ to!string(r) ~ "\n" ~ e);
+}
+
+void main()
```
Hackerpilot: What's wrong with `unittest` or the files with the `.args` extension in the test folder?

BBasile: - `.args` is not widely used, i didn't see it exists.
- `unittest` require a change in travis script that is a new test for them. (not a good idea to run them for each test of the standard suite).

stefan-koch-sociomantic:
@BBasile please propose the individual tests as a seperate PR.
Thanks.

PR's should be single feature-focused, unless there are cross-cutting concerns.

skl131313:
Tests that ensure a change or feature is working should be part of the same
PR, it doesn't make sense to put them into a different PR. That second PR
might not get created but this feature/change will already be in the repo.

BBasile:
There's no emergency, i can use the `.args` system. as said previouslt i didn't see it exists when i updated the PR this week-end. Let's tag this WIP.

README.md
BBasile: I cant change the default value because this would break the current behavior.

BBasile:
updated using `.args` files.

stefan-koch-sociomantic:
The code is not pretty, but I can go on with it.
LGTM

BBasile:
The same kind of PR will be necessary for the function declarations parameters.

PR #338, opened 2018-02-22
[trivial] Fix spacing in the README example

wilzbach:


WebFreak001:
while you are at it you might as well replace the tab directly above that line

PR #339, opened 2018-02-26
Fix formatting of empty files

stefan-koch-sociomantic:
With this fix we will no longer attempt to read empty files.
Which would cause a phobos exception due to a 0 size buffer being created.


ben-palmer-sociomantic:
LGTM

stefan-koch-sociomantic:
yes an empty file is valid d code according to grammar.

PR #342, opened 2018-03-01
Fix the issue #237

### LINKED ISSUE #237: Consecutive struct initializations confuses dfmt
``` d
struct A
{
    int x, y, z;
}

int main()
{
    int fun()
    {
        import std.stdio : writeln;
        import std.typecons : tuple;
        A a = { tuple(Variant(1))[0].get!int, tuple(Variant(2))[0].get!int, tuple(Variant(3))[0].get!int};
        A b = { tuple(Variant(1))[0].get!int, tuple(Variant(2))[0].get!int, tuple(Variant(3))[0].get!int};
        writeln(a);
    }
}

```

dfmt result:

``` d
struct A
{
    int x, y, z;
}
int main()
{
    int fun()
    {
        import std.stdio : writeln;
        import std.typecons : tuple;
        A a = {
            tuple(Variant(1))[0].get!int, tuple(Variant(2))[0].get!int,
                tuple(Variant(3))[0].get!int};
            A b = {
                tuple(Variant(1))[0].get!int, tuple(Variant(2))[0].get!int,
                    tuple(Variant(3))[0].get!int};
                writeln(a);
            }
        }

```

stefan-koch-sociomantic:
This fixes #237 by remembering whether we indented because of struct-initializers.
and then popping the indent specifically.

### DIFF
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -8,10 +8,19 @@ module dfmt.ast_info;
 import dparse.lexer;
 import dparse.ast;
 
+enum BraceIndentInfoFlags
+{
+    tempIndent = 1 << 0,
+}
+
 struct BraceIndentInfo
 {
     size_t startLocation;
     size_t endLocation;
+
+    uint flags;
+
+    uint beginIndentLevel;
 }
 
 /// AST information that is needed by the formatter.
@@ -39,6 +48,9 @@ struct ASTInformation
         sort(constructorDestructorLocations);
         sort(staticConstructorDestructorLocations);
         sort(sharedStaticConstructorDestructorLocations);
+
+        sort!((a,b) => a.endLocation < b.endLocation)
+            (indentInfoSortedByEndLocation);
     }
 
     /// Locations of end braces for struct bodies
@@ -92,7 +104,7 @@ struct ASTInformation
     /// Locations of constructor/destructor "this" tokens ?
     size_t[] constructorDestructorLocations;
 
-    BraceIndentInfo[] sortedByStartLocation;
+    BraceIndentInfo[] indentInfoSortedByEndLocation;
 }
 
 /// Collects information from the AST that is useful for the formatter
@@ -195,8 +207,12 @@ final class FormatVisitor : ASTVisitor
     {
         if (funcLit.functionBody !is null)
         {
-            astInformation.funLitStartLocations ~= funcLit.functionBody.blockStatement.startLocation;
-            astInformation.funLitEndLocations ~= funcLit.functionBody.blockStatement.endLocation;
+            const bs = funcLit.functionBody.blockStatement;
+
+            astInformation.funLitStartLocations ~= bs.startLocation;
+            astInformation.funLitEndLocations ~= bs.endLocation;
+            astInformation.indentInfoSortedByEndLocation ~=
+                BraceIndentInfo(bs.startLocation, bs.endLocation);
         }
         funcLit.accept(this);
     }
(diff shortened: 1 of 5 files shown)


stefan-koch-sociomantic:
this doesn't work to well because of `,`-wrap-Indents.

stefan-koch-sociomantic:
I am aware this does not look great, but making this work properly would mean a*major* re-factoring.

src/dfmt/formatter.d
```diff
@@ -850,7 +850,6 @@ private:
             }
             if (sBraceDepth > 0)
                 sBraceDepth--;
-
```
leandro-lucarella-sociomantic: Change unrelated to the commit, I would leave the line as it is since it doesn't really changes anything and just pollutes the diff.

src/dfmt/ast_info.d
```diff
@@ -8,6 +8,21 @@ module dfmt.ast_info;
 import dparse.lexer;
 import dparse.ast;
 
+enum LocationPairFlags
+{
+    TempIndent = 1 << 0,
+}
+
+struct LocationPair
```
Hackerpilot: A pair with four things in it?

src/dfmt/ast_info.d
```diff
@@ -85,6 +103,8 @@ struct ASTInformation
 
     /// Locations of constructor/destructor "this" tokens ?
     size_t[] constructorDestructorLocations;
+
+    LocationPair[] sortedByEndLocation;
```
Hackerpilot: This should probably be called something like `braceIndentInfo` because it contains information about the indentation of open/close brace pairs instead of information about sorting by end locations.

src/dfmt/formatter.d
```diff
@@ -803,15 +813,34 @@ private:
 
     void formatRightBrace()
     {
-        if (astInformation.structInitEndLocations.canFindIndex(tokens[index].index))
+        void popToBeginIndent(LocationPair locPair)
```
Hackerpilot: It looks like this is trying to solve the same problem that `indentToMostRecent` on the indent stack should be solving.

UplinkCoder: oh it was not oblivious from the code of from naming

UplinkCoder:
@Hackerpilot `indentToMostRecent` may not work if a non-struct-literal `{` has been pushed, at least I believe so.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic this currently blocks other work which needs the `BraceIdentInfo`

src/dfmt/ast_info.d
```diff
@@ -8,6 +8,21 @@ module dfmt.ast_info;
...
+{
+    TempIndent = 1 << 0,
+}
+
+struct BraceIndentInfo
+{
+    size_t startLocation;
+    size_t endLocation;
+
+    uint flags;
```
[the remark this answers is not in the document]
stefan-koch-sociomantic: because it's not guaranteed to be just one value but could use remaining bits for something else.

src/dfmt/formatter.d
```diff
@@ -803,15 +813,34 @@ private:
...
+            if (indentInfo.flags & BraceIndentInfoFlags.TempIndent)
+            {
+                popToBeginIndent(indentInfo);
+                simpleNewline();
+                indent();
+            }
             writeToken();
         }
-        else if (astInformation.funLitEndLocations.canFindIndex(tokens[index].index))
+        else if (astInformation.funLitEndLocations.canFindIndex(tokens[index].index, &pos))
```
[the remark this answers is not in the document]
stefan-koch-sociomantic: actually all of indices in here _have_ to be unique.
If they are not we really messed up something, I like having the assert.

PR #348, opened 2018-03-28
Deploy binaries to GitHub releases (Windows/Linux/MacOS)

wilzbach:
Basically a copy of https://github.com/dlang-community/D-Scanner/pull/573 which itself is a copy of https://github.com/dlang/dub/pull/1369

(I already configured the environment variables `GH_REPO_TOKEN` on Travis - it's a bot account and the same we already use for e.g. deploying libdparse's documentation)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,3 +43,55 @@ jobs:
               on:
                   tags: true # must be a git tag
                   repo: dlang-community/dfmt # must be a tag on dlang-community
+        - stage: GitHub Release
+          #if: tag IS present
+          d: ldc-1.8.0
+          os: linux
+          script: echo "Deploying to GitHub releases ..." && ./release.sh
+          deploy:
+            provider: releases
+            api_key: $GH_REPO_TOKEN
+            file_glob: true
+            file: bin/dfmt-*.tar.gz
+            skip_cleanup: true
+            on:
+              repo: dlang-community/dfmt
+              tags: true
+        - stage: GitHub Release
+          #if: tag IS present
+          d: ldc-1.8.0
+          os: osx
+          script: echo "Deploying to GitHub releases ..." && ./release.sh
+          deploy:
+            provider: releases
+            api_key: $GH_REPO_TOKEN
+            file_glob: true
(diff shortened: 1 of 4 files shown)


wilzbach:
> It seems that the .tar.tgz that was attached to the v0.5.0 release of D-Scanner has some issues. I had to manually chmod +x it,

That's weird. I didn't had to do this (and can't reproduce this).

> So it seems that we need to statically link Phobos when making the release executables.

This was my mistake. I build the v0.5.0 binary on my machine as there were still some problems with Travis. As it turns out that ArchLinux recently switched to pass `-link-defaultlib-shared` by default. The "official" LDC release binaries (i.e. the one used on Travis), don't have this problem and use static linking by default.
I just uploaded a fixed release archive.
LDC only has a `-static` flag which will result in linking in all system dependencies, but I'm not whether we want to do this either.
So I'm not sure what I can do about this in this PR.

Here's how the tarball would look like if built with the "official" LDC 1.8:

[dfmt-v0.5.2-linux-x86_64.tar.gz](https://github.com/dlang-community/dfmt/files/1870811/dfmt-v0.5.2-linux-x86_64.tar.gz)

```
ldd ./dfmt
	linux-vdso.so.1 (0x00007fff02fe0000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007fe8629e2000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007fe8627de000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fe8625c0000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007fe862274000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fe86205d000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fe861ca6000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007fe862e42000)
```

wilzbach:
(Added all changes that happened at the DScanner repo to this PR - including the build of the Windows binaries - directly on Travis)

PR #355, opened 2018-05-09
Better formatting for UFCS chains

### LINKED ISSUE #260: ufcs chains / pipelines
Breaking this off from https://github.com/Hackerpilot/dfmt/issues/7 as it's only one part of that.

This pattern is really very common in D:

``` D
auto blah = myRange
        .myCleverAlgorithm!(a => a.animal)
        .filter!"a.isButterfly"
        .each!spewFromDvdDrive;
```

and dfmt really mangles this sort of thing. I don't really know what algorithmic rules could be created that would make this work, but it would be great if there was something that at least caught common cases (`dfmt off/on` is fine, but obviously if a lot of code isn't being automatically formatted then that sucks. Ends up with a tension between getting things auto-formatted but not wanting so much fine-grained `dfmt off/on` visual noise).

Hackerpilot:
Fixes #260

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -4,7 +4,7 @@
     "targetType": "autodetect",
     "license": "BSL-1.0",
     "dependencies": {
-      "libdparse": "~>0.8.0-alpha.5"
+      "libdparse": "~>0.8.4"
     },
     "targetPath" : "bin/",
     "targetName" : "dfmt",
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -29,7 +29,8 @@ struct ASTInformation
     /// Sorts the arrays so that binary search will work on them
     void cleanup()
     {
-        import std.algorithm : sort;
+        import std.algorithm : sort, uniq;
+        import std.array : array;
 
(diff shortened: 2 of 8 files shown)


Hackerpilot:
CI failed because the dub repository hadn't updated. I've tried kicking it into re-running but haven't gotten Github to figure out that the build passes: https://travis-ci.org/dlang-community/dfmt/builds/376625013

BBasile:
> CI failed because the dub repository hadn't updated

It was but it's green now. I restarted the failed job manually.

> haven't gotten Github to figure out that the build passes

When the branch is pushed in the origin there are two builds and it was the other that failed. A setting on TravisCI can be deactivated to prevent this (the PUSH test and only keep the PR one) otherwise it's necessary to push on your own fork.

Hackerpilot:
CI issues aside, this is kind of a big change. Some of you might want to check out this branch and see what it does to your code.

stefan-koch-sociomantic:
@Hackerpilot @BBasile I am going to click the rebase button as soon as I can confirm it does not break ocean.

PR #360, opened 2018-06-14
Handle `do` as contract header and handle `body` as identifier

BBasile:


BBasile:
@stefan-koch-sociomantic  / @leandro-lucarella-sociomantic  Can you release ? I think i've made an error by incrementing dparse patch instead of minor version. I think you even don't have the choice.

BBasile:
Ping for the tag.

Issue #375, opened 2018-09-05
Update to dparse 0.9.6


BBasile:
Unfortunately with the FE 2.082 and the previous still not supported, the work to do has accumulated.

- expression-based contracts. (2.081)
- UDA for functions parameters. (2.082)
- enum member attributes. (2.082)

LaurentTreguier:
I was just looking into expression-based contracts.

BBasile:
If you want latest DCD in DLS i believe that 0.9.6 is necessary. There was an important fix related to scopes. (DCD requires >= 0.9.6 < 0.10.0)

LaurentTreguier:
Yep, I can't upgrade DCD or D-Scanner until DFMT also uses libdparse 0.9.x

BBasile:
done, even 0.9.8 now

PR #376, opened 2018-09-05
Support for DIP1009 (new contracts syntax), #375

LaurentTreguier:
I don't know if what I did is a correct way to do it (I've never really touched lexing and such before).
In `isBlockHeaderToken()` I added `in` because just like `out`, it can be a block header with classic contract syntax (maybe that should be in a separate commit ?).
In `isBlockHeader()`, I think changing `i + index >= tokens.length` to `i + index + 1 >= tokens.length` should not be a problem since even if the token is not outside of the range of `tokens`, if it's the very last one it shouldn't be a block header (if it is, then we'd be formatting invalid code anyway).

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1048,7 +1048,7 @@ private:
             else if (peekBackIsKeyword)
                 write(" ");
             writeToken();
-            if (!currentIs(tok!"(") && !currentIs(tok!"{"))
+            if (!currentIs(tok!"(") && !currentIs(tok!"{") && !currentIs(tok!"comment"))
                 write(" ");
             break;
         case tok!"try":
@@ -1090,7 +1090,7 @@ private:
                     tok!"}", tok!"=", tok!"&&", tok!"||") && !peekBackIsKeyword())
                 write(" ");
             writeToken();
-            if (!currentIs(tok!"(") && !currentIs(tok!"{"))
+            if (!currentIs(tok!"(") && !currentIs(tok!"{") && !currentIs(tok!"comment"))
                 write(" ");
             break;
         case tok!"case":
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1906,7 +1906,7 @@ const pure @safe @nogc:
     bool isBlockHeaderToken(IdType t)
     {
         return t == tok!"for" || t == tok!"foreach" || t == tok!"foreach_reverse"
-            || t == tok!"while" || t == tok!"if" || t == tok!"out"
+            || t == tok!"while" || t == tok!"if" || t == tok!"in"|| t == tok!"out"
             || t == tok!"do" || t == tok!"catch" || t == tok!"with"
             || t == tok!"synchronized" || t == tok!"scope";
     }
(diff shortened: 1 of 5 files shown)


LaurentTreguier:
I was way too hasty and forgot to run the tests...

BBasile:
You must add tests with the new syntax too.

LaurentTreguier:
Yes. I guess I have to start thinking before acting...

LaurentTreguier:
When I made this PR I had completely forgotten that obviously `out` blocks can have an argument too; so the only difference in this case is that block-based `out` contracts have a block following the argument...

So I have something that seems to be working (that doesn't break the tests, and with its own test files), however I had to make changes to libdparse again to resolve this problem: I added a `isExpression` property to the `InStatement` and `OutStatement` node types, since I didn't find a way to differentiate an expression-based contract from a block-based one in dfmt.

Do you see another way to do this ?

BBasile:
`out` followed by a left paren is always a contract. Isn't this enough to format a one line expression ?

LaurentTreguier:
`out` contracts followed by a left paren can be a block one : https://dlang.org/spec/contracts.html#pre_post_contracts

BBasile:
actually (sorry previous answer was wrong) you should be able to detect an expression based contract with the peek function family, e.g

```d
 if (peeIs(tok!"(", true) &&
     // out(;...)
    (peek2Is(tok!";", true) ||
    // out(a;...)
    (peek2Is(tok!"identifier", true) && peek2Is(tok!";", true)))){}   
```

BBasile:
edit: peek3Is

LaurentTreguier:
And this time I forgot about to update the git submodule. When I can get back on this the next push shall fix it for good, I swear !!

src/dfmt/formatter.d
```diff
@@ -1908,8 +1908,21 @@ const pure @safe @nogc:
...
+        const token = tokens[i + index];
+        const t = token.type;
+        bool isExpressionContract;
+
+        if (i + index + 3 < tokens.length)
+        {
+            IdType[3] toks;
+            foreach (j, ref tok; toks)
+                tok = tokens[i + index + j + 1].type;
+            isExpressionContract = (t == tok!"in" && toks[0] == tok!"(")
```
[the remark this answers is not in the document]
BBasile: test case

```d
out /*Major*/ ( /*Tom*/ result /*To ground control*/; true)
```

BBasile:
rebase against `master` please

BBasile:
You can approve this one https://github.com/dlang-community/dfmt/pull/383, after what you'll have to put the auto merge label and pull, rebase again to make the CI status green

tests/allman/dip1009.d.ref
```diff
@@ -0,0 +1,21 @@
...
+    assert(result == 0);
+}
+do
+{
+    return 0;
+}
+
+int bar(int arg)
+in(arg > 0)
+out(; true);
```
BBasile: I missed that: trailing `;` is invalid.

BBasile: i don't know how the tests site can work with this error

LaurentTreguier: As far as I understand the tests for dfmt concentrate on the formatted output, and not on the correctness of the input; putting this in libdparse's tests does make its test suite fail

PR #363, opened 2018-06-25
[WIP] Use automatic versioning

wilzbach:
Similar to https://github.com/dlang-community/DCD/pull/498 and the existing setup for DScanner.
This should set the version automatically for all three major build cases (build.bat, make, dub).

PR #379, opened 2018-09-06
Use automatic versioning

BBasile:
Like #363 but on 0.x.x


BBasile:
This makefile has become horrid.

BBasile:
It's green but the makefile require a cleanup. For example at some point `-Jbin` was added to the `INCLUDE_PATHS` variable.

BBasile:
The original was approved

veelo:
I fear this may be the cause of #407.

PR #381, opened 2018-09-06
Deprecate enforce ex (#377) for 0.8.x

BBasile:
picked from 0.x.x, sorry i 've get today that neptune propagate from bottom to top instead of the opposite.


BBasile:
backport ok ?

PR #382, opened 2018-09-06
Deprecate enforce ex (#377) for 0.7.x

BBasile:
Deprecate enforce ex
merged-on-behalf-of: BBasile <BBasile@users.noreply.github.com>


BBasile:
backport ok ?

WebFreak001:
it's the same functionality but only fixes a future breakage in newer compilers/phobos, so imo it could just be merged.

This would only work on dmd 2.067.1+ or ldc 0.16.0+ which are both pretty ancient (in the D world at least) What compiler version do these old dfmt versions require?

PR #383, opened 2018-09-08
deactivate testing with DMD nightly

BBasile:
The install script cannot be reached for now


BBasile:
See https://travis-ci.org/dlang-community/dfmt/builds/425584688?utm_source=github_status.

PR #386, opened 2018-09-19
Add test for enum member attributes and parameter attributes

BBasile:


BBasile:
ping @LaurentTreguier, this is indeed working without changes.

LaurentTreguier:
I've found one (albeit a bit far-fetched) weird case:
```d
enum E
{
deprecated
@
(
42
)
a
}

@(
42
)
int func(@
(
42
)
int a) { return a; }
```
Which produces:
```d
enum E
{
    deprecated @(42)
    a
}

@(42)
int func(@(42)
        int a)
{
    return a;
}
```

All new lines are removed, except right after the UDA when it uses parens.
The function parameter especially looks strange on its own line. (though I admit this is a kind of extreme test case)

BBasile:
formatAt is not adapted. For functions they are put on their own line. A change is required finally

BBasile:
enum will be easy (e.g ince enum is found, call into a new func called formatEnum) but function (which means also func literals) more tricky.

PR #393, opened 2018-10-07
Fix #286 - Weird indentation in else branch after wrapping expression

### LINKED ISSUE #286: weird indentation in else branch after wrapping expression
```d
void foo()
{
    if (true)
        enum vectorizeable = aLongExpressionThatCausesWrapping()
                && aLongExpressionThatCausesWrapping();
    else
                    enum vectorizeable = false;
}
```
Should be
```d
void foo()
{
    if (true)
        enum vectorizeable = aLongExpressionThatCausesWrapping()
                && aLongExpressionThatCausesWrapping();
    else
        enum vectorizeable = false;
}
```
instead

LaurentTreguier:
Popping wrap indents after end-of-expression semicolons fixes it.

### DIFF
--- a/tests/allman/issue0286.d.ref
+++ b/tests/allman/issue0286.d.ref
@@ -2,14 +2,14 @@ void foo()
 {
     if (true)
         enum vectorizeable = aLongExpressionThatCausesWrapping()
-                && aLongExpressionThatCausesWrapping();
+            && aLongExpressionThatCausesWrapping();
     else
         enum vectorizeable = false;
 
     if (true)
     {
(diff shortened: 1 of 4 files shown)


tests/allman/issue0286.d.ref
```diff
@@ -0,0 +1,16 @@
+void foo()
+{
+    if (true)
+        enum vectorizeable = aLongExpressionThatCausesWrapping()
+                && aLongExpressionThatCausesWrapping();
```
BBasile: Is the indentation supposed to be 8 spaces wider than the previous level ?

PR #395, opened 2018-10-08
Fix #372 - Improper indentation after single-line catch

### LINKED ISSUE #372: Improper formatting on this text
**What I expect:**

```
/// Entry to call the deflection, or error out.
void main(string[] args) {
    if (args.length > 1)
        try
            openUri(getRegistryInfo()["BrowserPath"], rewriteUri(args[1]));
        catch (Exception error)
            createErrorDialog(error);
    else
        createErrorDialog(new Exception("Expected one URI argument, recieved: \n" ~ args.to!string()));
}
```

**What I get:**

```
/// Entry to call the deflection, or error out.
void main(string[] args) {
    if (args.length > 1)
        try
            openUri(getRegistryInfo()["BrowserPath"], rewriteUri(args[1]));
        catch (Exception error)
            createErrorDialog(error);
    else
                createErrorDialog(new Exception("Expected one URI argument, recieved: \n" ~ args.to!string()));
}
```

Why does indent that second to last line so far?

LaurentTreguier:


BBasile:
thanks, nice work. I'm thinking about pushing 0.9.0, what do you think ? The recent changes related to attribs are unreleased.

LaurentTreguier:
That might be a good idea, the last changes have been piling up a bit.
A lot of these bug-tagged issues seems to be easily fixable; I think I'll continue scrolling through them in the coming days.

PR #411, opened 2018-11-14
Fix #407 - `dub run dfmt` keeps building dfmt anew each time

### LINKED ISSUE #407: `dub run dfmt` keeps building dfmt anew each time.
Something has regressed between version 0.8.2 and 0.8.3, which invalidates dub's cache it seems. Upon every invocation of `dub run dfmt -- mysource.d`, dub decides to rebuild dfmt anew, adding a good 10 seconds to every run. This makes it practically unusable in my case.

Probably a `preGenerateCommands` or `preBuildCommands` was added somewhere?

LaurentTreguier:
The reason dub rebuilds dfmt all the time is because `dubhash.txt` gets modified with the `preGenerateCommand`. By making this command only re-write `dubhash.txt` when the hash actually changed, then dfmt will not be rebuilt if no change has occured.

Also, I couldn't stand this huge single line in `dub.json` so I moved it in its own file instead of making it even bigger.

### DIFF
--- a/dubhash.d
+++ b/dubhash.d
@@ -1,0 +1,23 @@
+import std.algorithm;
+import std.ascii;
+import std.conv;
+import std.exception;
+import std.file;
+import std.path;
+import std.process;
+import std.range;
+import std.string;
+
+void main()
+{
+    auto dir = environment.get("DUB_PACKAGE_DIR");
+    auto hashFile = dir.buildPath("bin", "dubhash.txt");
+    auto gitVer = executeShell("git -C " ~ dir ~ " describe --tags");
+    auto ver = (gitVer.status == 0 ? gitVer.output.strip
+            : "v" ~ dir.dirName.baseName.findSplitAfter(
+                environment.get("DUB_ROOT_PACKAGE") ~ "-")[1]).ifThrown("0.0.0")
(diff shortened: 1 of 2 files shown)


LaurentTreguier:
Yep, I'll do that as well

dubhash.d
```diff
@@ -0,0 +1,23 @@
...
+import std.exception;
+import std.file;
+import std.path;
+import std.process;
+import std.range;
+import std.string;
+
+void main()
+{
+    auto dir = environment.get("DUB_PACKAGE_DIR");
```
WebFreak001: I love this change, but now that this is a separate file, there is a possibility a user runs this file outside of dub. Maybe this could check for dir.length and return if empty?

LaurentTreguier: Right now, if the variable doesn't exist, it throws an exception and exits anyway, so it won't try to write a file at some random place in the filesystem (if that's what you want to avoid)

WebFreak001: if that's the case then this is fine by me, the constant recompilations were getting really annyoing

LaurentTreguier: No kidding... dfmt + dscanner + dcd + actual package = 😩

PR #414, opened 2018-12-09
fix dub dependency build (fix #413)

### LINKED ISSUE #413: dub dependency broken
When trying to compile with dfmt ~master as dependency this error happens:

```
WARNING: A deprecated branch based version specification is used for the dependency dfmt. Please use numbered versions instead. Also note that you can still use the dub.selections.json file to override a certain dependency to use a branch instead.
Invalid source/import path: /home/webfreak/.dub/packages/dfmt-master/dfmt/bin
Invalid source/import path: /home/webfreak/.dub/packages/dfmt-master/dfmt/bin
Generating test runner configuration 'workspace-d-test-library' for 'library' (library).
Invalid source/import path: /home/webfreak/.dub/packages/dfmt-master/dfmt/bin
Running pre-generate commands for dfmt...
Error: module `dubhash` is in file 'dubhash.d' which cannot be read
import path[0] = .
import path[1] = /usr/include/dlang/dmd
Failed: ["/usr/bin/dmd", "-v", "-o-", "dubhash.d", "-I."]
Command failed with exit code 1: rdmd dubhash.d
```

WebFreak001:
run rdmd with `$PACKAGE_DIR` dub variable for an absolute path. The dubhash.d already uses the environment variable set by dub which tells it where the package is

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -15,6 +15,6 @@
         "built_with_dub"
     ],
     "preGenerateCommands" : [
-      "rdmd dubhash.d"
+      "rdmd $PACKAGE_DIR/dubhash.d"
     ]
 }


WebFreak001:
did you test on windows? I'm not 100% sure if the forward slash will also work on windows, but I am guessing it should work. If it does work I think this is mergable

LaurentTreguier:
Didn't think of that, but forward slashes have always worked on Windows for me... I'll still check just in case though

LaurentTreguier:
Doesn't seem to pose any problem. Forward slashes should be avoided for basic windows commands, for example `del bin/dubhash.txt` doesn't work because cmd.exe thinks `/dubhash.txt` is an option (even though it doesn't have a space before it...). But rdmd doesn't mind.

PR #438, opened 2019-03-17
upgrade dependencies

Basile-z:
To match DCD and D-Scanner dependencies but also to get latest dparse bugfixes.


Basile-z:
Wait DMD 2.085.1 before merging, see dlang-community/DCD/pull/590 for explanations.

PR #439, opened 2019-04-13
 make readme look nicer

dayllenger:
I got rid from that nasty horizontal scroll and polished few things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,33 +23,33 @@ By default, dfmt reads its input from **stdin** and writes to **stdout**.
 If a file name is specified on the command line, input will be read from the
 file instead, and output will be written to **stdout**.
 
-**dfmt** uses EditorConfig files for configuration. If you run **dfmt** on a
-source file it will look for .editorconfig files that apply to that source file.
-If no file is specified on the command line, **dfmt** will look for .editorconfig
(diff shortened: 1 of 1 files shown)


WebFreak001:
looks a bit more broken on the dub registry now but I think that should be fixed there

dayllenger:
Hell, certainly, our main issue is that we have almost no single mature library. Like what can be simpler than markdown? If we implement it once, it will stay there and help to many people for years. We need some kind of registry, in the spirit of Rusty *arewegameyet*, *arewewebyet*, etc., to know exactly what parts of ecosystem are stable and working, and what waits to be done.

Basile-z:
yeah current rendering is hilarious @WebFreak001. So  if this PR fixes the issue LGTM. maybe a tag is required  too.

Basile-z:
I've just pushed a tag, let's see how the dub page looks in two hours.

WebFreak001:
my screenshot was of a local dub instance how the PR will look on the dub registry. (so these screenshots weren't how it was looking before)

Before it looked like this:




so it's definitely not an improvement on the dub side, but I think on GitHub it does look nicer and dub should fix it instead of the dfmt README

Basile-z:
"whoops"

PR #442, opened 2019-05-13
Fix #361

### LINKED ISSUE #361: Unexpected empty new line if function header ends with a comment
The following code

```d
void foo() /**/
in
{
}
body
{
}
```

is formatted as

```d
void foo() /**/

in
{
}
body
{
}

```

kotet:


Basile-z:
Please put the title of the issue after the hashtag reference in the future, for example here it would have be :  `fix #361 - Unexpected empty new line if function header ends with a comment`. Despite of being the reporter I had no idea of what was the problem until I check it.

Issue #254, opened 2016-08-05
Reformatting objects completely messes up previous formatting


WebFreak001:
(edit note: this issue is not about broken struct formatting, it is about that dfmt should honor previous formatting of lines and not completely override it)

If I have a struct like this:

``` d
void main() {
    S s = {
        someStructMember1: 2,
        someStructMember2: 42,
        someStructMember3: null, // foobar
        someOtherMember1: objA,
        someOtherMember2: objB,
        someOtherMember3: 0,
        somethingMore: null,
        someFlagInThisStruct: -1
    };
}
```

it reformats it to this:

``` d
void main()
{
    S s = {
    someStructMember1:
        2, someStructMember2 : 42, someStructMember3 : null, // foobar
    someOtherMember1 : objA, someOtherMember2
            : objB, someOtherMember3 : 0, somethingMore : null, someFlagInThisStruct : -1};
    }
```

This formatting would _only_ make sense if this was a one-line struct before formatting like a small config struct or arguments for a function (and still for that the indentations and new lines are still super ugly).

However before formatting this was a definition where every variable had its own line and IMO dfmt should format it accordingly if my statement was multiple lines before formatting or not.

This results in really ugly code right now if you use this on large code files without `//dfmt off` and `//dfmt on` blocks. Also I am not a fan of these comments because they are used way too often to actually result in good looking code. The formatter should take away most work of making the code look good and not add more work.

Hackerpilot:
If I had to guess, the code that tries to figure out the difference between a struct literal and a function literal has messed up again.

WebFreak001:
fixed with #471 using `--keep_line_breaks true`

PR #446, opened 2019-07-30
Fix: Issue 256 "named member struct initialisers"

Sobaya007:


WebFreak001:
can you add more fields to the test to see how it wraps?

Sobaya007:
I added 2 cases but I don't know this is sufficient.

Any other cases are required?

WebFreak001:
I think there is still a small indentation bug there (extra spaces before every `:` after the first item) but it's already a lot better than what is currently in dfmt.

Can you try fixing the extra spaces issue? If not I think we should still get this merged soon as it's an improvement already.

(for reference the previous dfmt output for others)
```d
void foo()
{
    S s1 = {a:
    3};
    S s2 = {a:
    3, b : "test string"};
    S s3 = {a:
    3, b : "test string", c : {x:
    3.14, y : 3 + 4}};
}
```

Sobaya007:
Sorry I did not notice such an issue!
Is it OK?

Sobaya007:
Since I found a similar issue(#254), I also fixed test code and indentation bug.

WebFreak001:
ok looking good now, I didn't mind the previous test too btw because it has nested structs, which is a nice thing to test, so adding those again would be good.

Your current formatting will probably not be final, because it should probably wrap each property in a separate line for big blocks (like in big arrays) but that's for another PR.

Sobaya007:
I added the previous cases.

OK, then I'll fix the formatting for big structs and send another PR after.

src/dfmt/formatter.d
```diff
@@ -745,10 +749,13 @@ private:
         }
         else if (peekBackIs(tok!"identifier") && (peekBack2Is(tok!"{", true)
                 || peekBack2Is(tok!"}", true) || peekBack2Is(tok!";", true)
-                || peekBack2Is(tok!":", true)) && !(isBlockHeader(1) && !peekIs(tok!"if")))
+                || peekBack2Is(tok!":", true) || peekBack2Is(tok!",", true))
+                && !(isBlockHeader(1) && !peekIs(tok!"if")))
```
WebFreak001: this `&&` looks very confusing now, can you surround the 3 expressions with parentheses so they look like they belong together again?

src/dfmt/formatter.d
```diff
@@ -1551,6 +1558,12 @@ private:
                 immutable l = indents.indentToMostRecent(tok!"switch");
                 if (l != -1 && config.dfmt_align_switch_statements == OptionalBoolean.t)
                     indentLevel = l;
+                else if (astInformation.structInfoSortedByEndLocation
+                    .canFind!(st => st.startLocation < current.index && current.index < st.endLocation)) {
+                    immutable l2 = indents.indentToMostRecent(tok!"{");
+                    assert(l2 != -1);
```
WebFreak001: are you sure this l2 can't be -1? I would put in a reason why it should work as assert message (second string argument) so that in case this ever breaks it is possible to debug it and for others reading it to know why it is used without understanding all the code context.

src/dfmt/formatter.d
Sobaya007: What you mentioned is peekBack2Is(tok!",", true)), isBlockHeader(1) and peekIs(tok!"if")?

I cannot surround these because the first one is already surrounded with the previous terms.

WebFreak001:
```d
if (...
		peekBack2Is(tok!":", true) || /* add parens here: */ (peekBack2Is(tok!",", true))
		&& !(isBlockHeader(1) && !peekIs(tok!"if")))
```

WebFreak001: sorry - my issue, I misread it because it had too many parentheses already lol

Sobaya007: What about using staticMap instead?
```d
staticMap!(tok, "{","}",";",":",",")].any!((ulong token) => peekBack2Is(cast(IdType)token, true))
```

WebFreak001: that actually looks really readable then, but is the cast there needed? Maybe just using an array of `tok!""` wouldn't harm either and it would drop some work from the compiler. (because staticMap might be a bit of unnecessary compiler overhead here)

Sobaya007: I tried but I could not remove the cast from this code because the type of `tok`  is `ulong` and `IdType` cannot be `ulong`.
This is the same wheter or not I use staticMap.

src/dfmt/formatter.d
```diff
@@ -747,10 +747,10 @@ private:
...
             }
         }
-        else if (peekBackIs(tok!"identifier") && (peekBack2Is(tok!"{", true)
-                || peekBack2Is(tok!"}", true) || peekBack2Is(tok!";", true)
-                || peekBack2Is(tok!":", true) || peekBack2Is(tok!",", true))
-                && !(isBlockHeader(1) && !peekIs(tok!"if")))
+        else if (peekBackIs(tok!"identifier")
+                && [tok!"{", tok!"}", tok!";", tok!":", tok!","]
+                .any!((ptrdiff_t token) => peekBack2Is(cast(IdType)token, true))
+                && (!isBlockHeader(1) || peekIs(tok!"if")))
```
WebFreak001: this is unrelated but I think this is a good change (removes the double negation) so I approve of this

Sobaya007: Is there anything to do?

WebFreak001: no I think just waiting for other reviews now if there is anyone who would like to review

WebFreak001:
would merge in 2 days if nothing comes up

tests/allman/issue0256.d.ref
```diff
@@ -0,0 +1,11 @@
...
+{
+    S s1 = {a: 3};
+    S s2 = {a: 3, b: "test string"};
+    S s3 = {a: 3, b: "test string", c: {x: 3.14, y: 3 + 4}};
+    T t = {
+        someStructMember1: 2, someStructMember2: 42, someStructMember3: null, // foobar
+        someOtherMember1: objA, someOtherMember2: objB, someOtherMember3: 0,
+        somethingMore: null, someFlagInThisStruct: -1
+    };
+}
```
ZombineDev: Shouldn't this be:
```D
void main()
{
    S s1 = {a: 3};
    S s2 = {a: 3, b: "test string"};
    S s3 = {a: 3, b: "test string", c: {x: 3.14, y: 3 + 4}};
    T t =
    {
        someStructMember1: 2,
        someStructMember2: 42,
        someStructMember3: null, // foobar
        someOtherMember1: objA,
        someOtherMember2: objB,
        someOtherMember3: 0,
        somethingMore: null,
        someFlagInThisStruct: -1
    };
}
```

1. Struct initializer brace on a new line
2. User newlines and indentation preserved

WebFreak001: for now it's just an improvement over what we had before (completely broken indentation), so I don't think we quite need to implement the full formatting fix with this PR

Issue #234, opened 2016-03-01
How to build on Windows? *EDIT* Add DUB note to README


bruno-medeiros:
Is using `dub build` enough to build dfmt? 

I've used DUB on dfmt-0.4.5 and it seems to work fine, yet the README states one should use make to build dfmt (https://github.com/Hackerpilot/dfmt#building) , and, in the case of Windows there is also `build.bat` ? What's the story? I'm really hoping DUB is enough to build dfmt.

Hackerpilot:
The idea is that on Windows you can choose between the batch file or Dub, and on other operating systems you can use the makefile or Dub.

Hackerpilot:
The README doesn't mention Dub because I forgot to update it after adding Dub support.

bruno-medeiros:
Don't add it to the documentation _after_, add it _during_. :sunglasses:  That is, update documentation on the same commit as the new functionality that is being introduced. That's one of the big advantages of having the doc on the same repo as the source code. (One reason why all the documentation for my Github projects like DDT is in the Git repo, and not on the wiki. wiki is disabled in fact)
That said it's a hard habit to be consistent at, I still forget about it once in a while :sweat_smile:

PR #460, opened 2019-11-25
Fix trivial issues

WebFreak001:
Adds tests for what's already fixed

Also updates the README for #234 and updates the gen_expected.sh script for better usage.

note: I reproduced all open issues, these are all that are trivially fixed right now with example code with them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,18 +5,23 @@
 **dfmt** is beta quality. Make backups of your files or use source control
 when using the **--inplace** option.
 
-## Building
-### Using Make
-* Clone the repository
-* Run ```git submodule update --init --recursive``` in the dfmt directory
-* To compile with DMD, run ```make``` in the dfmt directory. To compile with
-  LDC, run ```make ldc``` instead. The generated binary will be placed in ```dfmt/bin/```.
+## Installation
 
 ### Installing with DUB
 
 ```sh
 > dub fetch --version='~master' dfmt && dub run dfmt -- -h
 ```
+
+### Building from source using Make
+* Clone the repository
+* Run ```git submodule update --init --recursive``` in the dfmt directory
+* To compile with DMD, run ```make``` in the dfmt directory. To compile with
+  LDC, run ```make ldc``` instead. The generated binary will be placed in ```dfmt/bin/```.
+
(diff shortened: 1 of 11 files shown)


tests/gen_expected.sh
```diff
@@ -1,10 +1,12 @@
...
+
 argsFile=$1.args
 if [ -e ${argsFile} ]; then
 	args=$(cat ${argsFile})
 fi
 echo "Args:" ${args}
-dfmt --brace_style=allman ${args} $1.d > allman/$1.d.ref
-dfmt --brace_style=otbs ${args} $1.d > otbs/$1.d.ref
+../bin/dfmt --brace_style=allman ${args} $1.d > allman/$1.d.ref
+../bin/dfmt --brace_style=otbs ${args} $1.d > otbs/$1.d.ref
```
wilzbach: Uhm should we put bin onto PATH instead?

WebFreak001: hm this is just run when the user makes new tests and wants to generate what dfmt currently outputs.

Before it uses the global dfmt which was not necessarily the newest dfmt. I think it's better if it uses the one in this repository path.

WebFreak001: any other idea? I think it's better than using the dfmt in PATH (because I don't think most users would put their development dfmt on PATH but instead a normal one for IDE usage)

Hackerpilot: This is probably a good change. I think that I originally wrote it this way because had a dfmt symlink in my path that pointed to my checkout of the dfmt code, so on my machines the dfmt on the path was always then one that was just compiled by the testing scripts.

README.md
```diff
@@ -5,19 +5,24 @@
...
 
 ### Installing with DUB
 
 ```sh
 > dub fetch --version='~master' dfmt && dub run dfmt -- -h
 ```
 
+### Building from source using Make
+* Clone the repository
+* Run ```git submodule update --init --recursive``` in the dfmt directory
```
PetarKirov: I'd suggest flattening these 2 steps into:
```
git clone --recurse-submodules https://github.com/dlang-community/dfmt
```

PR #464, opened 2020-01-03
Put space between a comment and an identifier

belka-ew:
Actual formatting:

```diff
-@nogc
-//
+@nogc//
 void foo();
```

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -448,7 +448,7 @@ private:
                 write(" ");
             else if (prevTokenEndLine == currTokenLine || (t == tok!")" && peekIs(tok!"{")))
                 write(" ");
-            else if (t == tok!"else")
+            else if (peekBackIsOneOf(false, tok!"else", tok!"identifier"))
                 write(" ");
             else if (canAddNewline || (peekIs(tok!"{") && t == tok!"}"))
                 newline();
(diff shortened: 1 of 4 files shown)


src/dfmt/formatter.d
```diff
@@ -448,7 +448,7 @@ private:
                 write(" ");
             else if (prevTokenEndLine == currTokenLine || (t == tok!")" && peekIs(tok!"{")))
                 write(" ");
-            else if (t == tok!"else")
+            else if (t == tok!"else" || t == tok!"identifier")
```
WebFreak001: can we change this to
```d
else if (t.among!(tok!"else", tok!"identifier"))
```
with
```d
import std.algorithm : among;
```
so future additions to this will be easier to write?

belka-ew: does `peekBackIsOne` the same? or do you prefer `among`?

WebFreak001: `peekBackIsOneOf` will work too, won't require an extra import then

belka-ew:
> does this also fix

No, it adds and additional space, as before as far as I can see.

WebFreak001:
ok that question doesn't matter then, would just prefer the `peekBackIsOneOf` call instead now

belka-ew:
One dmd build fails, it seems unrelated, but I can repush if needed.

WebFreak001:
I restarted that job

WebFreak001:
got that segmentation fault 3 times in a row now, will try again in a bit and otherwise merge too

belka-ew:
Weird. I had this on the first commit, repushed, and then the job succeeded.

PR #470, opened 2020-03-06
Fix Issue #469

### LINKED ISSUE #469: Invalid escape sequences truncate output and errorcode = success
example code (should be \ueee8 for valid unicode escape):

import std.stdio; void main() { writeln("\eee8Hello"); int a = 5; }

outputs:

main.d(0:0)[error]: Primary expression expected
main.d(0:0)[error]: Expected `}` instead of `EOF`
import std.stdio;

void main()
{
    writeln(!ERROR!

with error code = 0, which is painful when using Format on save option in IDEs because it can quietly remove huge parts of code.

Hackerpilot:
Check for lexer errors, and exit with a non-zero status if any are found. Fixes #469.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -16,7 +16,17 @@ import dfmt.tokens;
 import dfmt.wrapping;
 import std.array;
 
-void format(OutputRange)(string source_desc, ubyte[] buffer, OutputRange output,
+/**
+ * Formats the code contained in `buffer` into `output`.
+ * Params:
+ *     source_desc = A description of where `buffer` came from. Usually a file name.
+ *     buffer = The raw source code.
+ *     output = The output range that will have the formatted code written to it.
+ *     formatterConfig = Formatter configuration.
+ * Returns: `true` if the formatting succeeded, `false` of a lexing error. This
+ *     function can return `true` if parsing failed.
+ */
+bool format(OutputRange)(string source_desc, ubyte[] buffer, OutputRange output,
         Config* formatterConfig)
 {
     LexerConfig config;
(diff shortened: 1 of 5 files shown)


Hackerpilot:
Forgot about the dub build...

tests/test.sh
```diff
@@ -6,13 +6,22 @@ do
...
 		diff -u "${braceStyle}/${source}.ref" "${braceStyle}/${source}.out"
 	done
 done
+
+set +e
+
+for source in expected_failures/*.d
+do
+	if ../bin/dfmt "${source}" > /dev/null; then
+		exit 1
```
WebFreak001:
```suggestion
		echo "Expected failure on test ${source} but passed"
		exit 1
```

src/dfmt/formatter.d
```diff
@@ -33,11 +43,18 @@ void format(OutputRange)(string source_desc, ubyte[] buffer, OutputRange output,
     auto visitor = new FormatVisitor(&astInformation);
     visitor.visit(mod);
     astInformation.cleanup();
-    auto tokens = byToken(buffer, config, &cache).array();
+    auto tokenRange = byToken(buffer, config, &cache);
+    auto app = appender!(Token[])();
+    for (; !tokenRange.empty(); tokenRange.popFront())
+        app.put(tokenRange.front());
+    auto tokens = app.data;
```
WebFreak001: why no longer using `array()`?
```d
auto tokenRange = byToken(buffer, config, &cache);
auto tokens = tokenRange.array();
```
should work fine, no?

src/dfmt/main.d
```diff
@@ -211,14 +211,15 @@ else
                 else
                     break;
             }
-            format("stdin", buffer, output.lockingTextWriter(), &config);
+            return format("stdin", buffer, output.lockingTextWriter(), &config);
```
WebFreak001: pretty sure this will have the opposite effect of what is intended: return `true` = error code 1 (failure), return false = error code 0 (success)

src/dfmt/formatter.d
WebFreak001: nvm array would have save'd the struct and not emitted any errors then

src/dfmt/main.d
Hackerpilot: What's really dumb is that I got this right in the other branch of the code...

PR #473, opened 2020-03-23
keep_line_breaks: multi-line tokens and argument list break

belka-ew:
--keep_line_breaks only (diffs are the result without this PR).

Fixes line breaks in some argument lists:

```diff
     auto x = func(argument_1_1, argument_1_2,
-            this.argument_2_1, this.argument_2_2,
+            this.argument_2_1,
+            this.argument_2_2,
             argument_3_1, argument_3_2);
 ```

and multi-line tokens:

```diff
 
 `
     <html>
     </html>
-`.format!"%s";
+`
+    .format!"%s";
 
```

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1574,13 +1574,6 @@ private:
             newline();
             regenLineBreakHints(index - 1);
         }
-        else if (!peekIs(tok!"}") && (linebreakHints.canFind(index)
-                || (linebreakHints.length == 0 && currentLineLength > config.max_line_length)))
-        {
-            pushWrapIndent();
-            writeToken();
-            newline();
-        }
         else if (config.dfmt_keep_line_breaks == OptionalBoolean.t)
         {
             const commaLine = tokens[index].line;
@@ -1598,6 +1591,13 @@ private:
                     newline();
                 }
             }
+        }
+        else if (!peekIs(tok!"}") && (linebreakHints.canFind(index)
+                || (linebreakHints.length == 0 && currentLineLength > config.max_line_length)))
+        {
+            pushWrapIndent();
+            writeToken();
+            newline();
         }
         else
         {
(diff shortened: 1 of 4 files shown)


src/dfmt/formatter.d
```diff
@@ -2148,13 +2148,18 @@ const pure @safe @nogc:
...
+    bool onNextLine() @nogc pure @safe
     {
         import dfmt.editorconfig : OptionalBoolean;
+        import std.algorithm.searching : count;
+
+        if (config.dfmt_keep_line_breaks == OptionalBoolean.f || index <= 0)
+        {
+            return false;
+        }
+        const tokenEndLine = tokens[index - 1].line + tokens[index - 1].text.count('\n');
```
WebFreak001: keep this nothrow because it is used like a property everywhere, use `.text.representation.count('\n')` to avoid unicode exceptions

src/dfmt/formatter.d
```diff
@@ -2148,13 +2148,18 @@ const pure @safe @nogc:
...
 
-        return config.dfmt_keep_line_breaks == OptionalBoolean.t
-            && index > 0
-            && tokens[index - 1].line < tokens[index].line;
+        return tokenEndLine < tokens[index].line;
```
WebFreak001: can you add a comment explaining what is being done in these 2 lines? `tokenEndLine` isn't the most intuitive variable name to interpret

WebFreak001: 👍 very nice, explains the need for `\n` too. I'm usually more of a fan of renaming variables to act as a kind of documentation, but that's a bit hard to do here.

belka-ew: Yeah, I couldn't think of a better name. If you have an idea, I'll rename it.

belka-ew:
Fixed a typo in the comment, `token[index - 1]` instead of `tokens[index - 1]`.

src/dfmt/formatter.d
WebFreak001: how about previousTokenEndLineNo

belka-ew: Done

PR #475, opened 2020-04-02
Fix #474

### LINKED ISSUE #474: Assertion when trying to format "scope (exit)"
```
$ ./bin/dfmt --version
v0.12.0-beta.1-3-g1964f80-debug
$ echo 'scope (exit)' | ./bin/dfmt 
stdin(1:1)[error]: Basic type, type constructor, symbol, `typeof`, `__traits`, `__vector` or `mixin` expected
core.exception.AssertError@src/dfmt/formatter.d(1986): Assertion failure
----------------
??:? _d_assertp [0x55fd5646addd]
src/dfmt/formatter.d:1986 const pure nothrow ref @nogc @safe const(std.experimental.lexer.TokenStructure!(ubyte, "\x0a    string comment;\x0a    string trailingComment;\x0a\x0a    int opCmp(size_t i) const pure nothrow @safe {\x0a        if (index < i) return -1;\x0a        if (index > i) return 1;\x0a        return 0;\x0a    }\x0a\x0a    int opCmp(ref const typeof(this) other) const pure nothrow @safe {\x0a        return opCmp(other.index);\x0a    }\x0a").TokenStructure) dfmt.formatter.TokenFormatter!(std.stdio.File.LockingTextWriter).TokenFormatter.current() [0x55fd563b50f8]
src/dfmt/formatter.d:1111 void dfmt.formatter.TokenFormatter!(std.stdio.File.LockingTextWriter).TokenFormatter.formatBlockHeader() [0x55fd563b14c2]
src/dfmt/formatter.d:271 void dfmt.formatter.TokenFormatter!(std.stdio.File.LockingTextWriter).TokenFormatter.formatStep() [0x55fd563ad1f2]
src/dfmt/formatter.d:131 void dfmt.formatter.TokenFormatter!(std.stdio.File.LockingTextWriter).TokenFormatter.format() [0x55fd563acbd2]
src/dfmt/formatter.d:56 bool dfmt.formatter.format!(std.stdio.File.LockingTextWriter).format(immutable(char)[], ubyte[], std.stdio.File.LockingTextWriter, dfmt.config.Config*) [0x55fd563ac468]
src/dfmt/main.d:219 _Dmain [0x55fd563bb86a]
```

This happens in earlier versions too (I encountered it in v0.10.1)

WebFreak001:
also creates a property to check for token existence called `hasCurrent` returning `index < tokens.length` (replaced previous occurences)

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1093,46 +1093,51 @@ private:
             indents.push(current.type);
         }
         writeToken();
+
         if (currentIs(tok!"("))
         {
             write(" ");
             writeParens(false);
         }
-        if (currentIs(tok!"switch") || (currentIs(tok!"final") && peekIs(tok!"switch")))
-            write(" ");
-        else if (currentIs(tok!"comment"))
-            formatStep();
-        else if (!shouldPushIndent)
+
+        if (hasCurrent)
         {
-            if (!currentIs(tok!"{") && !currentIs(tok!";"))
+            if (currentIs(tok!"switch") || (currentIs(tok!"final") && peekIs(tok!"switch")))
                 write(" ");
-        }
-        else if (!currentIs(tok!"{") && !currentIs(tok!";") && !currentIs(tok!"in") &&
-            !currentIs(tok!"out") && !currentIs(tok!"do") && current.text != "body")
-        {
-            newline();
-        }
-        else if (currentIs(tok!"{") && indents.topAre(tok!"static", tok!"if"))
-        {
-            // Hacks to format braced vs non-braced static if declarations.
-            indents.pop();
-            indents.pop();
-            indents.push(tok!"if");
-            formatLeftBrace();
-        }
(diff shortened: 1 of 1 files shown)


src/dfmt/formatter.d
```diff
@@ -1868,7 +1873,8 @@ private:
             else
                 formatStep();
         }
-        while (index < tokens.length && parenDepth > 0);
+        // TODO: obviously getting stuck here?
+        while (hasCurrent && parenDepth > 0);
```
WebFreak001, line 1877: not quite sure how this isn't getting stuck before? Both are only getters, not modifying the state?

Hackerpilot: Are you saying that you think that this is an infinite loop? `formatStep` and `writeToken` calls cause later calls to `hasCurrent` to return a different value.

WebFreak001: ohh nvm I didn't notice this was a do-while loop, I thought it was an empty while loop

WebFreak001:
cool dlang-bot works again!

PR #492, opened 2020-06-10
Fix #490

### LINKED ISSUE #490: keep_line_breaks causes extra line added after comment
```
% cat dfmtmess.d
void foo()
{
    afdsafds
        .asdf // blah
        .flub;
}
% dfmt --keep_line_breaks true dfmtmess.d
void foo()
{
    afdsafds
        .asdf // blah
        
        .flub;
}
```

Hackerpilot:


WebFreak001:
this breaks all other cases:
```d
void foo() // hello
{ // world
    // ok
    writeln("hi"); // hi!
} // done
//finish
```
becomes
```d
void foo() // hello{ // world// okwriteln("hi"); // hi!} // done//finish
```

Hackerpilot:
I'm kind of thinking that the larger issue here is that our regression tests are not thorough enough.

Hackerpilot:
To further elaborate, a while back I made the tester run all the test files with each brace style. I'm starting to think that this keep_line_breaks option is just as important, or maybe more important.

WebFreak001:
I agree, I think having the keep_line_breaks work great will feel like a new formatter for a lot of people once widely used.

How about also adding a basic test which tests if dfmt accidentally breaks a program in any way? With reproducible builds this could simply be: compile a large project, run dfmt on the large project, compile it again and see if the executable is still the same (so we don't accidentally break stuff like with these comments or previously with that !ERROR! token that occurred on parsing errors)

Hackerpilot:
I have some local changes to the `test.sh` and `gen_expected.sh` scripts that run the four combinations of brace style and keep_newline. The downside to this is that the number of "expected" files is twice as large as before, and also that nobody has looked at their outputs yet. The "allman" and "otbs" reference files were created and reviewed one-at-a-time as each bug was fixed, and the new variants caused by turning on the keep_line_breaks option would have to be reviewed.

WebFreak001:
superseded by #514

PR #499, opened 2020-08-03
Break line on multiline strings

belka-ew:
2 Bugs:

- `chooseLineBreakTokens` expects to get tokens of a single line. But since a string with newlines is a single token, the newlines inside it weren't acknowledged. Further `chooseLineBreakTokens` calculates the length of the tokens by using `tokens.map!(a => tokenLength(a)).sum()`. `tokenLength()` takes only the length of the string literal before the newline (which is actually correct if no other tokens forllow the string).
- `currentLineLength += cast(uint) o; // where o is the length of the string after the last newline`. `currentLineLength` should be just `o` since the length is just the length on the string on the new line without the length of the tokens on the previous line.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1622,16 +1622,21 @@ private:
     {
         import std.range : assumeSorted;
         import std.algorithm.comparison : min;
-        import std.algorithm.searching : countUntil;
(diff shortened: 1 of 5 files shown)


belka-ew:
No auto-merge label :(

Issue #501, opened 2020-08-21
Weird formatting for spaces in associative arrays


arkark:
Input:

```d
void main()
{
    auto aa = ["aaa": 1, "bbb":2];
}
```

Actual output with default options (so, `dfmt_space_before_aa_colon` is `false`):

```d
void main()
{
    auto aa = ["aaa" : 1, "bbb" : 2];
}
```

Expected output (I think):

```d
void main()
{
    auto aa = ["aaa": 1, "bbb": 2];
}
```

---

https://github.com/dlang-community/dfmt/blob/58b2c6ebc620af1351821509b435c821cd175d90/src/dfmt/formatter.d#L841

I don't know why `indents.topDetails.breakEveryItem` is checked here.

Issue #485, opened 2020-04-27
hashmap with symbols not recognized/indented wrongly


FeepingCreature:
Consider the following code:
```
void main()
{
    int a;
    int[int] hashmap = [a : a, a : a, a : a];
}
```
Valid, and DMD accepts it. However, dfmt turns it into the very strange:
```
void main()
{
    int a;
    int[int] hashmap = [a : a, a:
        a, a:
        a];
}
```

Note that the first key-value has the correct spacing, but the others don't. Does it think it's doing labels?

belka-ew:
Same as #462.

WebFreak001:
#462 is caused by being in a delegate of that specific size I believe

Issue #462, opened 2019-12-25
Weird AA formatting


0xEAB:
> `serve-d v0.6.0 with workspace-d v3.5.0`

The AA gets formatted in an unexpected way.

```d
bool asdf(const string owner, const string mail) @safe
{
    requestHTTP(url, (scope HTTPClientRequest request) {
        request.writeFormBody([owner: owner, mail:
                mail]);
    }, (scope HTTPClientResponse response) {});

    return true;
}
```

WebFreak001:
this is using dfmt 0.11.0

PR #502, opened 2020-08-23
Print colon in AA before identifiers

belka-ew:
- #485 
- #462 
- #501


belka-ew:
@Hackerpilot can you please close the referred issues?

PR #505, opened 2020-09-01
Keep line break after function attributes

### LINKED ISSUE #504: 'deprecated' is moved to the same line even with `--keep-line-breaks=true`.
Repro:

```
(echo 'deprecated("foo")'; echo 'void test() { }') |bin/dfmt --keep_line_breaks=true
```

Output:

```
deprecated("foo") void test()
{
}
```

Expected:

```
deprecated("foo")
void test()
{
}
```

Note that it does not happen with

```
deprecated
void test() { }
```

belka-ew:
Fixes #504.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -713,13 +713,13 @@ private:
         {
             writeToken();
             if (spaceAfterParens || parenDepth > 0)
-                write(" ");
+                writeSpace();
         }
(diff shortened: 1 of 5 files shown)


PetarKirov (approved):
Looks good to me

PetarKirov:
I think @HackerPilot wants to have a look as well, so I'll let him merge when he's ready.

PR #507, opened 2020-09-10
Keep line breaks in UFCS chains

### LINKED ISSUE #503: Chain break with template arguments
```diff
 string main()
 {
-    return duration.total!"seconds".to!string;
+    return duration.total!"seconds"
+        .to!string;
 }
```

It doesn't seem to happen if the chain has only one `!`.

belka-ew:
Fixes #503.

I looked at the code and dfmt makes best effort to recognize what is a UFCS chain. I'm not sure what is the correct behaviour. But --keep_line_breaks adding an additional newline is definitely a bug, so I fixed only that part.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -1401,7 +1401,8 @@ private:
             break;
         case tok!".":
             regenLineBreakHintsIfNecessary(index);
-            immutable bool ufcsWrap = astInformation.ufcsHintLocations.canFindIndex(current.index);
+            immutable bool ufcsWrap = config.dfmt_keep_line_breaks == OptionalBoolean.f
+                    && astInformation.ufcsHintLocations.canFindIndex(current.index);
             if (ufcsWrap || linebreakHints.canFind(index) || onNextLine
                     || (linebreakHints.length == 0 && currentLineLength + nextTokenLength() > config.max_line_length))
             {
--- a/tests/allman/issue0503.d.ref
+++ b/tests/allman/issue0503.d.ref
@@ -1,0 +1,15 @@
+string f()
+{
+    return duration.total!"seconds".to!string;
+}
+
+string g()
+{
+    return duration.total!"seconds"().to!string;
+}
+
+string h()
+{
+    return duration.total!"seconds"().to!string.to!string.to!string.to!string.to!string.to!string
+        .to!string.to!string.to!string;
+}
--- a/tests/issue0503.args
+++ b/tests/issue0503.args
@@ -1,0 +1,1 @@
+--keep_line_breaks=true
--- a/tests/issue0503.d
+++ b/tests/issue0503.d
@@ -1,0 +1,14 @@
+string f()
+{
+    return duration.total!"seconds".to!string;
+}
+
+string g()
+{
+    return duration.total!"seconds"().to!string;
+}
+
+string h()
+{
+    return duration.total!"seconds"().to!string.to!string.to!string.to!string.to!string.to!string.to!string.to!string.to!string;
+}
--- a/tests/otbs/issue0503.d.ref
+++ b/tests/otbs/issue0503.d.ref
@@ -1,0 +1,12 @@
+string f() {
+    return duration.total!"seconds".to!string;
+}
+
+string g() {
+    return duration.total!"seconds"().to!string;
+}
+
+string h() {
+    return duration.total!"seconds"().to!string.to!string.to!string.to!string.to!string.to!string
+        .to!string.to!string.to!string;
+}


src/dfmt/formatter.d
```diff
@@ -1401,12 +1401,14 @@ private:
         case tok!".":
             regenLineBreakHintsIfNecessary(index);
             immutable bool ufcsWrap = astInformation.ufcsHintLocations.canFindIndex(current.index);
```
Hackerpilot, line 1403: I was able to get your tests to pass by reverting all these changes and replacing them with
```d
immutable bool ufcsWrap = config.dfmt_keep_line_breaks == OptionalBoolean.f
        && astInformation.ufcsHintLocations.canFindIndex(current.index);
```

Hackerpilot:
> But --keep_line_breaks adding an additional newline is definitely a bug, so I fixed only that part.

The documentation states that the option means "Keep existing line breaks if these don't violate other formatting rules." Which I think, so far, has meant that existing line breaks will be maintained, but that new ones can be added if there are lines that are too long. If the formatter is breaking lines that aren't actually too long, then that's a different problem. And in that case, checking the value of the keep_line_breaks config option is not the right way to do things.

src/dfmt/formatter.d
belka-ew: I wasn't aware I broke long line wrapping. I applied your suggestion and added a test for a long line. dfmt still breaks the line at a strange position: not 80 characters, not 120, but 100. But I think it is a different thing.

belka-ew:
@Hackerpilot, is it good to go now?

belka-ew:
Anybody?

belka-ew:
Three green bottles standing on the wall
Three green bottles standing very tall
And if one green bottle should accidentally fall
There'll be two green bottles standing on the wall

belka-ew:
Two green bottles standing on the wall
Two green bottles standing very tall
And if one green bottle should accidentally fall
There'll be one green bottle standing on the wall

PetarKirov (approved):
LGTM
@Hackerpilot @WebFreak001?

WebFreak001 (approved):

Hackerpilot:
I've been putting off looking at this for two reasons.
1) The first is that nobody seems to have actually read my original comment on this pull request. The `keep_line_breaks` option is used to *preserve*, not *prevent*, line breaking. Adding a check of this configuration option here is not the right thing to do.
2) The second is that the real solution to this problem is to define a correct way to decide if the code in question is a function call pipeline or just a member function call. I don't know how to do this. The best that we have is the code in the ast_info module that generates the UFCS hints, and it's pretty clear that this is not good enough.

WebFreak001 (changes requested):
@Hackerpilot you are right, I have simply looked at the tests here and thought it would be good enough. But when turning off keep_line_breaks, the `h()` function is formatted with each `.to!string` on a separate line, which is imo much more what we want to have.

belka-ew:
1. > But when turning off keep_line_breaks, the h() function is formatted with each .to!string on a separate line, which is imo much more what we want to have.
  
    It is not different from how this is handled currently. Please note that the pull request doesn't change the behaviour without `-keep_line_breaks`. Everything it does is disabling the automatically generated line breaks **with** `--keep_line_breaks`.

2. This change doesn't *prevent* all line breaks, it only prevents smart line breaks inserted based on the AST information to format the UFCS chains differently. It doesn't prevent the line break if the line is too long. I specifically added a test with `h()` which proves my claim and breaks the line with `--keep_line_breaks` if the line gets too long. And it is exactly what `keep_line_breaks` is supposed to do, it says: Let me decide manually where to break the line if it's not too long.

3. I was smart enough to remove my original repository (and I apologize). I recreated it and rebased the branch on master, but it doesn't links to this pull request automatically, so I probably need to recreate the pull request.

WebFreak001:
> It is not different from how this is handled currently. Please note that the pull request doesn't change the behaviour without --keep_line_breaks. Everything it does is disabling the automatically generated line breaks with --keep_line_breaks.

Yes the problem here is that these automatically generated line breaks that help in readability are missing with `--keep_line_breaks` here.

As Hackerpilot said:

> The keep_line_breaks option is used to preserve, not prevent, line breaking. Adding a check of this configuration option here is not the right thing to do.

and I agree with that.

Users use the auto-formatter to not need to manually insert new lines and indentation, but instead let dfmt automatically do so.

This is what dfmt does right now:

```d
string f()
{
    return duration.total!"seconds"
        .to!string;
}

string g()
{
    return duration.total!"seconds"().to!string;
}

string h()
{
    return duration.total!"seconds"().to!string
        .to!string
        .to!string
        .to!string
        .to!string
        .to!string
        .to!string
        .to!string
        .to!string;
}
```

And I think that's much better than changing that behavior to the here suggested

```d
string f()
{
    return duration.total!"seconds".to!string;
}

string g()
{
    return duration.total!"seconds"().to!string;
}

string h()
{
    return duration.total!"seconds"().to!string.to!string.to!string.to!string.to!string.to!string
        .to!string.to!string.to!string;
}
```

FeepingCreature:
Disclaimer: I work for Funkwerk.

I understand what you're saying about `--keep-line-breaks`. But I think it's underappreciated that the line breaks that dfmt creates at the moment are *aesthetically terrible.* Would it be acceptable to introduce another flag, such as `--dont-reflow-property-chains`?

I mean, I'm a dfmt user as well, and I think this formatting rule makes the appearance of the formatted code gratuitously worse.

WebFreak001:
I think if we make this a separate flags it would be fine

PR #523, opened 2020-11-18
Document Visual Studio integration.

veelo:
How dfmt can be configured as an external tool in Visual Studio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,12 @@ when using the **--inplace** option.
 * run `dub build --build=release`, optionally with `--compiler=ldc2`
 
 ## Using
-By default, dfmt reads its input from **stdin** and writes to **stdout**.
+By default, **dfmt** reads its input from **stdin** and writes to **stdout**.
 If a file name is specified on the command line, input will be read from the
 file instead, and output will be written to **stdout**.
+
+Several editor and IDE plug-ins exist with support for **dfmt**. Visual Studio
+can be configured to use **dfmt** as [an external tool](#visual-studio-integration).
 
 **dfmt** uses [EditorConfig](http://editorconfig.org/) files for configuration. If you run **dfmt** on a
 source file it will look for *.editorconfig* files that apply to that source file.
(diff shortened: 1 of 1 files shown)


Imperatorn:
Does it work with VS?

README.md
```diff
@@ -121,3 +124,15 @@ dfmt_keep_line_breaks | `true`, **`false`** | Keep existing line breaks if these
...
+**dfmt** can be configured as an external tool in Visual Studio (2017 and 2019). Choose [Tools]->[External Tools...], click [Add] and fill in the following fields:
+
+|                   |                                     |
+|-------------------|-------------------------------------|
+|Title:             | `d&fmt`                             |
+|Command:           | `C:\D\dmd2\windows\bin\dub.exe`     |
+|Arguments:         | `run dfmt -- --inplace $(ItemPath)` |
+|Initial Directory: | `$(ItemDir)`                        |
+
+Now Alt+T Alt+F will reformat the current file on disk. If you forgot to save the file first, VS will warn that the file changed on disk and ask if you want to discard unsaved changes; Reply No, save and try again.
```
WebFreak001, line 138: this keyboard shortcut (Alt-F) might be broken by localization. In German VS it works (text is `Ex&tras`) but I don't know if it's so convenient in all languages. You should add a note here that what this keyboard shortcut is doing is actually just navigating through the top menu by keyboard.

README.md
```diff
@@ -121,3 +124,15 @@ dfmt_keep_line_breaks | `true`, **`false`** | Keep existing line breaks if these
 * Braces - `{` and `}`
 * Brackets - `[` and `]`
 * Parenthesis / Parens  - `(` and `)`
+
+## Visual Studio integration
+**dfmt** can be configured as an external tool in Visual Studio (2017 and 2019). Choose [Tools]->[External Tools...], click [Add] and fill in the following fields:
```
WebFreak001, line 129: as a hint it might make sense to say Tools is Alt-T here

WebFreak001 (changes requested):

README.md
WebFreak001: also you don't need the Alt for the F key (even when not invoking the Tools menu using the keyboard)

Imperatorn: Ok, isn't the standard keybinding Ctrl + K + D? Could dfmt be registered to use that?

WebFreak001: at least not using this recommended way here, which is just running dfmt as external tool. I think using some VS extension it should be possible.

veelo:
Good point, thanks for the review.

In the meantime, dfmt has been integrated into VisualD, which is a better integration than this trick. I'm considering to just close this PR. Thoughts?

WebFreak001:
sounds good, that will contain better integration anyway.

Issue #530, opened 2021-03-08
Error when running installation command


preetpalS:
When you follow the installation instructions on the README, you get an error.

```
C:\Random\Path>dub fetch --version='~master' dfmt && dub run dfmt -- -h
The '--version' parameter was deprecated, use dfmt@'~master'. Please update your scripts.
Invalid SemVer format: '~master'

C:\Random\Path>
```

preetpalS:
Fixed with #531.

PR #531, opened 2021-03-08
Fix installation instructions on README.md

preetpalS:
Fixes Issue #530.

Also modified single quotes to be double quotes so command runs on Windows command line. Tested on Ubuntu 20.04 and Windows 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ when using the **--inplace** option.
 ### Installing with DUB
 
 ```sh
-> dub fetch --version='~master' dfmt && dub run dfmt -- -h
+> dub fetch dfmt@"~master" && dub run dfmt -- -h
 ```
 
 ### Building from source using Make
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ when using the **--inplace** option.
 ### Installing with DUB
 
 ```sh
-> dub fetch --version='~master' dfmt && dub run dfmt -- -h
+> dub fetch dfmt@"~master" && dub run dfmt -- -h
```
PetarKirov, line 13: Since a couple of versions, Dub will automatically fetch a package, if it doesn't exist, so we can skip the `fetch` part. Also I believe the `~master` suggestion is outdated - some time ago there was a big gap between the last official release and `master`, but I think it's not the case anymore.

```suggestion
> dub run dfmt -- -h
```

preetpalS: I committed the suggestion, it's a more straightforward way to install and run dfmt.

PetarKirov (approved):

PetarKirov:
Thanks @preetpalS!

PR #534, opened 2021-04-16
Add Kernighan & Ritchie style brace formatting

s-ludwig:
The original K&R style behaves like BraceStyle.otbs, except for function definitions that use Allman style.

See https://en.wikipedia.org/wiki/Indentation_style#K&R_style


s-ludwig:
Should be noted that I've taken the "otbs" tests and adjusted them by hand to fit the "knr" style.

s-ludwig:
Any chance of someone having a look at this? Would be nice to be able to use dfmt as a tool for contributions to vibe.d related projects.

Note that the actual change is pretty small, it's just a lot of test cases added.

tests/knr/2d_arrays.d.ref
```diff
@@ -0,0 +1,37 @@
+unittest {
```
WebFreak001, line 1: unittests are functions, so the brace should be on a separate line

s-ludwig: The idea with having function signatures on a separate line is to make them stand out while visually scanning the file. This doesn't really apply to unit test "functions".

src/dfmt/ast_info.d
```diff
@@ -200,6 +206,15 @@ final class FormatVisitor : ASTVisitor
         destructor.accept(this);
     }
 
+    override void visit (const FunctionBody functionBody)
+    {
+        if (auto bd = functionBody.specifiedFunctionBody)
+        {
+            astInformation.funBodyLocations ~= bd.blockStatement.startLocation;
```
WebFreak001, line 213: there could be a null dereference here if bd.blockStatement is null.

However with the current grammar this case is impossible, it could only be introduced in the future. I don't think it's very likely that this will happen to blockStatement (as shortened statement got a separate AST name), so I am fine with merging without checking for null here

WebFreak001 (approved):

WebFreak001:
I think in general it would be better if all of these states were tweakable separately, but would merge this one for now anyway. Would like to know what @Hackerpilot thinks before merging too though.

s-ludwig:
Agreed that separate switches would be preferable (IMO), but I gathered that the project vision was to be more opinionated rather than supporting customization. @Hackerpilot doesn't seem to be engaged much with open-source projects lately, though, so the question is how/if we can continue with this.

WebFreak001:
would merge today unless he appears then

PR #535, opened 2021-05-19
Add single_indent option

belka-ew:
Since the DStyle proposal to indent continuation lines by two tabs instead of
one was rejected (see https://github.com/dlang/dlang.org/pull/2766), and it is
the only behaviour dfmt supports, I suggest to add an option to indent
everything with a single tab. So the option changes the indentation inside the
parens.

Double indentation is useful to visually emphasize a code block after an
argument list or if-, while-conditions and so on. For example

```d
if (true
  && true)
  doSomething();
```
vs

```d
if (true
        & true)
    doSomething();
```

The indentation in the second case makes it clear where the condition ends.

Some style guides enforce:

- curly braces even in single line if-statements
- curly braces on the new line

In  the case of a such style guide, the problem disappears because there is
always a new line between the condition and the code block:

```d
if (true
    && true)
{
    doSomething();
}
```

It also eliminates the need to think about aligning the code with mixed parens,
brackets and braces (e.g. `([])`), just indent everything with one tab.


WebFreak001:
requesting additional review from Hackerpilot because the indentation engine is black magic to me

belka-ew:
Can we get this futher? Last week I've tested this option on a bunch of Funkwerk projects, hundreds of thousands of lines of code, and it works as expected with a few exceptions caused by existing indentation bugs.

Issue #494, opened 2020-07-02
Very long chained calls in array literals are indented weirdly


FeepingCreature:
Got:
```
foo([
        line1,
        value_line2_bla_bla_bla.propertyCallBlaBlaBla(a, b, c)
        .propertyCallBlaBlaBla(a, b, c, d).propertyCallBlaBlaBla(a, b, c)
        .propertyCallBlaBlaBla(a, b, c).value,
        ]);
```
Expected:
```
foo([
        line1,
        value_line2_bla_bla_bla.propertyCallBlaBlaBla(a, b, c)
            .propertyCallBlaBlaBla(a, b, c, d).propertyCallBlaBlaBla(a, b, c)
            .propertyCallBlaBlaBla(a, b, c).value,
]);
```

Two independent issues here is that the array literal closer and call closer aren't indented correct for *either* an array literal or a function call, and the properties aren't indented correctly which may be related to https://github.com/dlang-community/dfmt/issues/484 (but note that I don't pass any arguments in this example.)

Without the array literal, the closing paren is formatted correctly- but the indentation is still wrong.

Compare (with `--keep_line_breaks`):
```
int a = [
];
int b = foo(
    
);
int c = foo([
        ]);
```

PR #537, opened 2021-06-11
Fix double array indentation inside parens

belka-ew:
Fixes partly #494. If an array is inside parens, it gets the double identation of the parens and its own indentation. Instead the array indentation should override the parens, that indents the array elements correctly and the closing `])` as well.


PetarKirov (approved):
Ah those test case finally look correct, thanks! It's nice that the fix was this simple.

PR #540, opened 2021-07-28
Upgrade libdparse to latest version

RazvanN7:
This upgrade contains:

- alias assign recognition from the parser
- a memory corruption bug fix


WebFreak001 (approved):
because dfmt mainly uses just the tokenizer and doesn't rely too much on the AST I think bumping to <1.0.0 is ok for this

PR #541, opened 2021-08-03
fix the help for the config option

aminya:
The code uses `config`, but in the help it is shown as config_dir


PetarKirov (approved):
Thanks!

PR #547, opened 2022-02-13
Update dub.json to remove dependency on `rdmd`

skyline131313:


dub.json
```diff
@@ -15,6 +15,6 @@
         "built_with_dub"
     ],
     "preGenerateCommands" : [
```
WebFreak001, line 17:
```suggestion
    "preBuildCommands" : [
```

WebFreak001 (changes requested):
this impacts the compile time of every project depending on dfmt as a package, as preGenerateCommands is ran on every invocation. (not just when rebuilding a dependency) It might be possible to use `preBuildCommands` to generate it though, which will only impact the compile time once and then dub keeps track of the cache and doesn't run it again.

WebFreak001:
this project is missing CI, I will make a PR on D-Scanner to test this out first

WebFreak001 (approved):
will merge now without CI tests as it worked out fine in the D-Scanner repo: https://github.com/dlang-community/D-Scanner/pull/857

PR #550, opened 2022-05-23
Use specified line endings in multiline comments.

### LINKED ISSUE #228: end_of_line=crlf produces lf in multiline comments
Regular code has correct - crlf endings.
Tested on windows, v0.4.5

veelo:
fix #228


tests/issue0228.args
```diff
@@ -0,0 +1 @@
+--end_of_lin=crlf
```
WebFreak001, line 1:
```suggestion
--end_of_line=crlf
```

WebFreak001 (changes requested):
not sure if the tests folder is the right place for this, as git may change the line endings on different OSes or settings.

veelo:
> not sure if the tests folder is the right place for this, as git may change the line endings on different OSes or settings.

Good point. I'll remove the files from the repository and generate them instead.

> You would need to add replace code when outputting comments and strings (which will exhibit the same behavior) to fix this issue.

I think https://github.com/dlang-community/dfmt/pull/550/files#diff-f5d86cddb20837cd3a9f154e16b9a929d7b7d133ce71a5022127e5ec30d94a00 does the right thing. `eolString` is the platform-dependent standard line ending, or the specified one with `--end_of_line`.

`test.d` from #551 correctly detects #228, and this fixes it (barring your point above).

veelo:
> `eolString` is the platform-dependent standard line ending [...]

I take that back. It currently is LF by default, see #552.

veelo:
> not sure if the tests folder is the right place for this, as git may change the line endings on different OSes or settings.

If everything works as advertised, this shouldn't be happening courtesy `.gitattributes` that I added.

I had to rewrite this slightly to work properly for all kinds of line endings in the input.

src/dfmt/formatter.d
```diff
@@ -1834,12 +1836,7 @@ private:
         }
         else
         {
-            // You know what's awesome? Windows can't handle its own line
-            // endings correctly.
-            version (Windows)
-                output.put(current.text.replace("\r", ""));
-            else
-                output.put(current.text);
+            output.put(current.text.lineSplitter.joiner(eolString));
```
veelo, line 1839: For a moment I thought this would omit the trailing newline. But `dfmt` always ends the output with a newline even if the input doesn't, so t's good. #551 will detect if this ever changes.

WebFreak001: instead of doing the short and easy, but memory intensive way, how about adding a helper function that first checks if there are non-conforming line-endings and then runs the lineSplitter + appends it all on it?

WebFreak001: nvm I just realized these are ranges and should be fine as is

WebFreak001 (approved):

veelo:
Thanks.

PR #551, opened 2022-05-24
Port `test.sh` to platform independent D.

veelo:
This executes also way faster by not spawning GNU diff (at least on Windows).


veelo:
This script simply ignores differences in line ending, unless `--end_of_line` is one of the arguments to dfmt.

tests/test.d
```diff
@@ -0,0 +1,55 @@
...
+    foreach (braceStyle; ["allman", "otbs", "knr"])
+    {
+        foreach (entry; dirEntries(".", "*.d", SpanMode.shallow).filter!(e => e.baseName(".d") != thisExePath.baseName(".exe")))
+        {
+            const source = entry.baseName;
+            const argsFile = source.stripExtension ~ ".args";
+            string[] args = argsFile.exists ? readText(argsFile).splitter!isWhite.filter!(a => a.length).array : [];
+            writeln(dfmt, " --brace_style=", braceStyle, " ", args.map!(a => a ~ " ").join, source);
+            auto output = File(buildPath(braceStyle, source ~ ".out"), "w+");
+            int dfmt_result = spawnProcess([dfmt, "--brace_style=" ~ braceStyle] ~ args ~ [source],
```
maxhaton, line 22: `format`? It's easier to do logging when you avoid building process arguments in place using `~s`

veelo: Thanks, I have reworked the logging, amongst other edits.

veelo:
I am keeping this in draft for the moment because on Windows `dfmt` seems to generate LF line endings by default, which is unexpected. If that is indeed the case, this script should probably not ignore line endings like is does now.

veelo:
I had it backwards yesterday night: the script should be correct in it's current state for the current `dfmt` behaviour. Once #552 is addressed, the script can be simplified to always compare line endings.

I intentionally did not touch `test.sh`, as devs on Posix might want to test both versions for a while to make sure I didn't miss anything.

This should be ready for review now.

maxhaton:
Seems OK to me.

tests/test.d
```diff
@@ -0,0 +1,104 @@
...
+        foreach (entry; dirEntries(".", "*.d", SpanMode.shallow).
+                            filter!(e => e.baseName(".d") != thisExePath.baseName(".exe")))
+        {
+            const source = entry.baseName;
+            const outFileName = buildPath(braceStyle, source ~ ".out");
+            const refFileName = buildPath(braceStyle, source ~ ".ref");
+            const argsFile = source.stripExtension ~ ".args";
+            const dfmtCommand = 
+                [dfmt, "--brace_style=" ~ braceStyle] ~ 
+                (argsFile.exists ? readText(argsFile).splitter!isWhite.filter!(a => a.length).array : []) ~
```
WebFreak001, line 27:
```suggestion
                (argsFile.exists ? readText(argsFile).split : []) ~
```
the default split(ter) does it all for us (splits across multiple whitespace, ignores leading and trailing whitespace - will never produce an empty string in the result array)

tests/test.d
```diff
@@ -0,0 +1,104 @@
...
+version (Windows)
+    enum dfmt = `..\bin\dfmt.exe`;
+else
+    enum dfmt = `../bin/dfmt`;
+
+int main()
+{
+    foreach (braceStyle; ["allman", "otbs", "knr"])
+        foreach (entry; dirEntries(".", "*.d", SpanMode.shallow).
+                            filter!(e => e.baseName(".d") != thisExePath.baseName(".exe")))
```
WebFreak001, line 19: I think it would be better to hardcode `test.d` here as it doesn't necessarily need to be run with the same executable name, e.g. it may not match with a temporary executable.

Also can you move the . from above to this line (`.filter`)?

WebFreak001 (commented):
I think it might make sense to force an --end_of_line parameter to all dfmt calls, so all files are always tested with the platform native EOL string (if not changed to anything else)

tests/test.d
veelo: Yeah I was trying to accomplish that, but overread the details of the documentation. Thanks.

veelo:
Can we get this merged please? I have a PR for #552 ready that assumes this.

maxhaton (approved):

veelo:
Thanks, that was quick!

PR #553, opened 2022-05-29
Retain line ending by default.

### LINKED ISSUE #552: `end_of_line` defaults to LF, even on Windows.
`dfmt` produces non-native line endings on Windows, unless explicitly instructed with `--end_of_line=crlf`. This is unfortunate, as explicitly stating `--end_of_line` in scripts makes it difficult to maintain cross platform compatibility.

Most editors on Windows will produce CRLF line endings by default, and even Git will by default convert files to CRLF on checkout. Most text files will therefore have CRLF endings on Windows, and users may be surprised to find their files having LF endings after a run through `dfmt`.

There was a long discussion by the folks of `editorconfig` about [whether `--end_of_line` should support a `native` setting](https://github.com/editorconfig/editorconfig/issues/226), and whether that should be the default. It was decided that this was not a good idea, and that `unset` is the desired default, meaning to leave the default up to the editor.

But, since `dfmt` often introduces line breaks, it needs some kind of default. And LF is not the right default in every case, IMO.

The obvious solution would be to make CRLF the default on Windows, and LF everywhere else. However, as reported in the above linked thread, some people configure Git to check out all files using LF also on Windows, and they would not want `dfmt` to change line endings away from their preference.

The most versatile solution would probably be to inspect the first *n* lines of input and see what the dominating line endings are there, and use that as default for the entire output.

veelo:
If no `--end_of_line` was specified, find the first line ending in the input and use that for the whole output.

All tests include checking for differences in line endings.

Closes #552.


WebFreak001 (commented):
it seems the enum extension should be automatically supported, can you test it also?

veelo:
What is the enum extension?

WebFreak001:
sorry, I meant it looks like `--end_of_line=automatic` should just work now if we use the std.getopt or std.conv methods with the enum and that it should be documented and tested.

I thought I added a suggestion comment to the review as well but it seems like I didn't.

veelo:
Oh that. The main reason I didn't do that is because then the set of options would be different from [EditorConfig options](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#end_of_line). Before, we had the **`lf`** default, now we don't have a documented default just like EditorConfig. When an EditorConfig has no `end_of_line` defined then it is up to the editor to decide, and our equivalent is that it is up to `dfmt` to decide, which now takes a hint from the input. So I see the `automatic` enum member more as an implementation detail than a user visible feature.

Second reason is that is doesn't make much sense to me to supply the default value for an option on the command line.

I'll add a note to the readme what `dfmt` does when `end_of_line` is undefined.

WebFreak001:
for IDE tools / integrations it might make sense to know about the special value, so having it documented would be useful there I think.

src/dfmt/editorconfig.d
```diff
@@ -43,7 +43,8 @@ enum EOL : ubyte
     unspecified,
     lf,
     cr,
-    crlf
+    crlf,
+    automatic
```
WebFreak001, line 47: can we rename this to `_automatic` if it's just an internal value to make sure people don't use it?

veelo: My first take on this was indeed using `unspecified`, but that messed with line length calculations, where a line ending in `\r\n` started to break differently from the same line when it ended in `\n`.

`unspecified` is used when merging two configurations. By adding an additional enum member the line length issues disappeared.

WebFreak001: ok, makes sense. Yes I think also renaming it to _unspecified makes sense.

veelo:
Before now I didn't realize that `automatic` showed up for `dfmt --help`. Now it doesn't.

veelo:
This is good to go as far as I am concerned.

src/dfmt/formatter.d
```diff
@@ -121,8 +123,23 @@ struct TokenFormatter(OutputRange)
...
+                const firstCR = cast(uint) rawSource.countUntil("\r");
+                if (firstCR < cast(uint) rawSource.countUntil("\n"))
+                {
+                    if (firstCR == rawSource.countUntil("\r\n"))
+                        this.eolString = "\r\n";
+                    else
+                        this.eolString = "\r";
+                }
+                else
+                    this.eolString = "\n";
```
WebFreak001, line 141: isolate this logic into its own function pls, constructor might not be a good place for application logic.

WebFreak001 (commented):
LGTM other than this minor thing, can you also rebase onto master to trigger the CI with latest code?

veelo:
I am not seeing any CI. Is that just me?

Issue #435, opened 2019-02-17
dfmt v0.10.0-beta.1 array formatting


LaurentTreguier:
I ran dfmt v0.10.0-beta.1 on dls' code base to see how it changes. I always use dfmt to format it, so it's currently formatted with dfmt v0.9.0, and any change when formatting with a newer version should correspond to a change since v0.9.0.
Some observations:
- Ternary operators are better formatted (at least in the cases I've seen, it looks slightly better)
  This is caused by 6e4136a353f737ead16bd74636efd6a8c0d63898; since it's in the middle of commits about arrays and AA's it looks like it's a side effect (correct me if that's wrong). Example:
  ```diff
       override void visit(const UnionDeclaration dec)
       {
  -        visitSymbol(dec, SymbolKind.interface_, true, dec.structBody is null ? 0
  -                : dec.structBody.endLocation);
  +        visitSymbol(dec, SymbolKind.interface_, true, dec.structBody is null
  +                ? 0 : dec.structBody.endLocation);
       }
  ```
- When a line is long and contains array/AA literals, they are formatted on multiple lines somewhat unnecessarily. Examples:
  ```diff
   unittest()
   {
       // [...]
  -    assert(convertFromJSON!(int[string])(JSONValue([16, 42])) == ["0" : 16, "1" : 42]);
  +    assert(convertFromJSON!(int[string])(JSONValue([16, 42])) == [
  +            "0": 16,
  +            "1": 42
  +            ]);
   }
  ```
  ```diff
       if (!SymbolTool.instance.workspacesFilesUris.canFind!sameFile(uri))
       {
  -        send(TextDocument.publishDiagnostics, new PublishDiagnosticsParams(uri, []));
  +        send(TextDocument.publishDiagnostics, new PublishDiagnosticsParams(uri, [
  +                ]));
       }
  ```
- Large array and AA declarations are perfectly formatted however, as intended, and don't need `//dfmt off` and `//dfmt on` anymore

Full diff with the changes that I've found: [dfmt.diff.txt](https://github.com/dlang-community/dfmt/files/2872881/dfmt.diff.txt)

WebFreak001:
maybe some additional heuristics should be added (like for example if the array fits in one single line after wrapping)

I think in general more cases look better now than before, but these minor bugs are slightly buggin me.

```
diff --git a/source/dls/protocol/jsonrpc.d b/source/dls/protocol/jsonrpc.d
index d0a7a44..ac22838 100644
[...]

PR #554, opened 2022-06-03
Fix array literal indentation in foreach

belka-ew:
This is a more generic fix for the array indentation problems in parens. It was already fixed for array literals in function arguments, but there are other places it still didn't work. e.g. `foreach` statements.

Here is the old formatting with `--keep_line_breaks`:

```d
    foreach (x; [
        1,
        2,
        3,
        ])
    {
    }
```

With this change the formatting is:

```d
    foreach (x; [
        1,
        2,
        3,
    ])
    {
    }
```

It also fixes the formatting of the last example in this issue https://github.com/dlang-community/dfmt/issues/435.

And I changed `test.sh` to ignore `test.d`, because currently `test.sh` is broken.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -15,7 +15,7 @@ import dfmt.indentation;
 import dfmt.tokens;
 import dfmt.wrapping;
 import std.array;
-import std.algorithm.comparison : among;
+import std.algorithm.comparison : among, max;
 
 /**
  * Formats the code contained in `buffer` into `output`.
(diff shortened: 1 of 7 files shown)


maxhaton (approved):
LGTM, not very familiar with this code though

WebFreak001 (commented):
```d
void f()
{
    foreach (x; foo!(map!([
            1,
            2,
            3,
        ])))
    {
    }
```
still gets formatted like this, intended?

belka-ew:
> still gets formatted like this, intended?

No. Fixed and tests added.

PR #556, opened 2022-06-13
Change rdmd to $DC -run

### LINKED ISSUE #472: Dfmt dub packages relies on deprecated tool rdmd
DMD has now on windows a 64 version of the compiler. The deprecated tool `rdmd` is only available in the `bin` folder but not in the `bin64` folder. Therefore you likely get the error message `rdmd" is not recognized` during dub build. Maybe we can switch to a dub single package here instead of making use of the deprecated `rdmd` tool?

WebFreak001:
fix #472

Analogous to https://github.com/dlang-community/D-Scanner/pull/857

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -14,7 +14,7 @@
     "versions" : [
         "built_with_dub"
     ],
-    "preGenerateCommands" : [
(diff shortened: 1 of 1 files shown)


WebFreak001:
duplicate of #547

PR #560, opened 2022-07-12
CI

veelo:
Add GitHub action.

### DIFF
--- a/.github/workflows/d.yml
+++ b/.github/workflows/d.yml
@@ -1,0 +1,34 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+name: D
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


veelo:
The action in action: https://github.com/veelo/dfmt/runs/7299705939?check_suite_focus=true

WebFreak001 (commented):
this does not seem to be based on https://github.com/dlang-community/dfmt/blob/master/.travis.sh

I think it might be better to port the travis CI config as it currently does:

- build dfmt both with minimum supported libdparse and maximum supported libdparse
- test the makefile building
- run the test.sh script

veelo:
Indeed. But if it's already there, why isn't it in use? I don't really care how things are done, I just observed there was no CI.

> * run the test.sh script

That is Posix-only.

rikkimax:
> Indeed. But if it's already there, why isn't it in use? I don't really care how things are done, I just observed there was no CI.

https://forum.dlang.org/post/sfueznovjwywbnzhptha@forum.dlang.org

veelo:
I see.

Personally I am not planning on using more time on this, I think this PR is better than nothing, and can be extended to adopt more from `.travis.sh` by anyone who wants to. Whatever is decided, I do think it should use `test.d` as this PR does instead of `test.sh`, as `test.d` is platform independent and faster also.

I'll leave this PR open and see what happens, my main objective is to move forward with #553.

WebFreak001:
ok sounds good, I guess the old files are there to be picked up again later as well

PR #564, opened 2022-08-07
Fix else indentation after array literal

belka-ew:
It is for bug introducted by me, where the `else` get incorrect indentation if `if` has an `[]` in the condition. For example:

```d
    if (true)
    {
        if (args[0])
        {
        }
    else
        {
        }
    }
```

(`else` tries to find the indentation of the last `if` but it is removed from the indentation stack).

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -203,7 +203,7 @@ private:
 
     /// Tracks paren depth on a single line. This information can be used to
     /// indent array literals inside parens, since arrays are indented only once
-    /// and paren indentation is ignored.line breaks and "[" reset the counter.
+    /// and paren indentation is ignored. Line breaks and "[" reset the counter.
     int parenDepthOnLine;
 
     string eolStringFromInput() const
@@ -726,7 +726,6 @@ private:
             {
                 indents.pop();
             }
-            indents.popTempIndents();
         }
         parenDepthOnLine = 0;
     }
@@ -1703,8 +1702,6 @@ private:
         import std.algorithm : max, canFind;
         import dfmt.editorconfig : OptionalBoolean;
 
-        parenDepthOnLine = 0;
-
         if (currentIs(tok!"comment") && index > 0 && current.line == tokenEndLine(tokens[index - 1]))
             return;
 
@@ -1822,7 +1819,19 @@ private:
                 if (indents.topIs(tok!"]"))
                 {
                     indents.pop();
-                    indentLevel = indents.indentLevel;
+                }
+                // Find the initial indentation of constructs like "if" and
+                // "foreach" without removing them from the stack, since they
+                // still can be used later to indent "else".
+                auto savedIndents = IndentStack(config);
+                while (indents.length >= 0 && indents.topIsTemp) {
+                    savedIndents.push(indents.top, indents.topDetails);
+                    indents.pop;
+                }
+                indentLevel = indents.indentLevel;
+                while (savedIndents.length > 0) {
+                    indents.push(savedIndents.top, savedIndents.topDetails);
+                    savedIndents.pop;
                 }
             }
             else if (astInformation.attributeDeclarationLines.canFindIndex(current.line))
@@ -1853,6 +1862,7 @@ private:
             }
             indent();
         }
+        parenDepthOnLine = 0;
     }
 
     void write(string str)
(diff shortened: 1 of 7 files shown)


src/dfmt/formatter.d
```diff
@@ -726,7 +730,13 @@ private:
             {
                 indents.pop();
             }
-            indents.popTempIndents();
+            if (indents.length > 0 && indents.topIsTemp()) {
+                savedIndents ~= IndentStack(config);
+            }
+            while (indents.length > 0 && indents.topIsTemp()) {
+                savedIndents.back.push(indents.top, indents.topDetails);
```
WebFreak001, line 737: `.back.push`?

WebFreak001 (changes requested):
storing an array of stacks kind of seems weird. When exactly was the issue introduced and can we solve this by further adjusting that change?

belka-ew:
> When exactly was the issue introduced and can we solve this by further adjusting that change?

Here: https://github.com/dlang-community/dfmt/pull/554, d862d8aef1b46289be1fc7cf80809b70a734e8fa

I tried to find a better solution but came up with this. Probably I could look again at it, maybe I get a fresh idea.

The problem here is that `else` is indented same as `if`. The formatter when it sees an `else` looks for the last `if` on the indentation stack and puts the same amount of spaces as for `if`. My commit above removes the `if`-indentation from the stack because it breaks the array literal indentation.

There are actually a few constructs, that are normally indented the same, independently of braces before. One of the is array literals:

```d
if ((((([
    // One level indentation here, even if I'm in `if` or after braces.
])))))
```

Another one is probably multiline lambdas, that have a lot of indentation problems currently in arguments and other constructs

```d
if (f((x, y) => {
    // One indentation level here.
}));
```

So indentation stack is cleared for such constructs. But after the construct, other code can follow that should be indented and there is a need to restore the indentation stack to indent the code after such thing.

Therefore I came up with this array of indentation stacks, where I can save any amount of indentations (recursively) and restore them after a specific construct ends.

src/dfmt/formatter.d
belka-ew: A new element in the array is reserved in the condition before (`savedIndents ~= IndentStack(config);`) and that pushes the indentation to this element.

belka-ew:
Well after some testing I've found a different solution; the tests are left unchanged. I kept the first commit for comparision, but can do a squash.

WebFreak001:
feel free to squash yourself, would otherwise squash merge

WebFreak001 (changes requested):
Can you add tests for
```d
foreach (x; y)
	if (foo)
	{
	}
	else
	{
	}
```
and
```d
if (a)
	if (b)
	{
	}
	else
	{
	}
```

belka-ew:
Sure, done.

src/dfmt/indentation.d
```diff
@@ -278,6 +278,12 @@ private:
 
                 immutable currentIsNonWrapTemp = !details[i].wrap
                     && details[i].temp && arr[i] != tok!")" && arr[i] != tok!"!";
+
+                if (currentIsNonWrapTemp && arr[i + 1] == tok!"]")
```
WebFreak001, line 282: `arr[i + 1] == tok!"]"` looks super random here (and possible range violation?), can you explain what exactly it does, maybe in a comment?

otherwise the PR LGTM I think

WebFreak001: hm shouldn't `[` use a different indentation (temp? wrap?) instead of doing it on the `]`? What about nested arrays with this approach?

belka-ew: Works for tests like:

```d
private void selectMember(Args...)(Args args)
{
    if (true)
    {
        if (args[
            [
                0,
            ],
            [
                2,
            ],
        ])
        {
        }
        else
        {
        }
    }
}
```

For nested arrays this rule doesn't apply, because a nested array is a different sequence of indentation tokens, its `]` followed by another `]`.

Not sure about implementing it as a different kind of token. The documentation for temp states:

> temporary indentation which get's reverted when a block starts generally true for all tokens except ), {, case, @

Array indentation isn't temporary, it just shouldn't be summed up with the brace indentation.

For wrap:

> generally true for all operators except {, case, @, ], (, )

Wrapping is as far as I understand more a tool to split long sequences of operators or UFCS chains into multiple lines. Array is an item list, it seems to be a bit different.

WebFreak001 (approved):

PR #569, opened 2022-10-05
Fix #568, don't output empty files without error

### LINKED ISSUE #568: dfmt erase the file when it fails to format it
If a source file is not formatted correctly then dfmt sometimes erases the content of the file
 
File sample
```
### This file should not be erased just because it is not formatted correctly
```

The .editconfig used
```
# top-most EditorConfig file
root = true

# D source
[*.{d,di}]
dfmt_brace_style = stroustrup
dfmt_split_operator_at_line_end = true
indent_size = 4
soft_max_line_length = 80
dfmt_soft_max_line_length = 120
dfmt_keep_line_breaks = true
# dfmt_compact_labeled_statements = true
# conditional_newline_indent = true
# dfmt_template_constraint_style = conditional_newline
```

WebFreak001:
Now when an error in formatting happens, it never outputs anything and doesn't override the file when working inplace.

Additionally dfmt is no longer able to fail in the middle of a file, as now we first write everything to a buffer and only if everything was successful, that buffer is printed to stdout or written to the inplace file.

This should also guard against segfaults with inplace file formatting erasing parts of the file, as well as the user thinking it was successful, even though dfmt didn't finish properly.


tests/test.d
```diff
@@ -29,39 +29,55 @@ int main()
...
+            return ret;
+    }
+
     writeln("All tests succeeded.");
     return 0;
 }
+
+int diff(string refFileName, string outFileName)
+{
+    const outText = outFileName.readText;
```
maxhaton, line 58: Kingobird has a Myers diff impl in D if you need it

WebFreak001: I kept the original implementation from the test script, we can consider to switch it out in another PR

belka-ew (approved):
Looks good. But as far as I can see the test doesn't test the actual bug, it just tests that dfmt fails on invalid files.

WebFreak001:
the test checks that the file stays unchanged on error here: https://github.com/dlang-community/dfmt/pull/569/files#diff-8fe9cc3c7ea77894b9a69b9f03ff2b4ecd4ec8ac2ff190e00066af46f8c40ad2R48

belka-ew:
You're right, missed that.

PR #570, opened 2022-10-05
always add potential spaces after `@attr`

WebFreak001:
this uses a new `queueSpace` function, which inserts a space before writing the next token, unless we already had a newline inserted. This also fixes cases with double spaces.

This queueSpace function makes it much more resilient to grammar changes because we can put it in a lot more spaces without worrying about potentially inserting two spaces. It also catches more cases with keep_new_lines as every inserted space checks for if it should keep the existing new line at the current token.

### DIFF
--- a/src/dfmt/formatter.d
+++ b/src/dfmt/formatter.d
@@ -198,6 +198,11 @@ private:
     /// True if the next "this" should have a space behind it
     bool thisSpace;
 
+    /// True if the next write call should first write a space, unless it is a
+    /// new line or other whitespace, in which case no further space is inserted.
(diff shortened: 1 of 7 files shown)


WebFreak001:
got some issues with additionaly spaces inserted after/before comments now

linkrope:
> got some issues with additionaly spaces inserted after/before comments now

@WebFreak001:  do you plan to fix the issues in the near future?
Maybe we should have a release once again: should this release be delayed?

WebFreak001:
I think I'd actually just say we rather shouldn't merge this and instead fix the issues when they come up, since it's just how dfmt works now

PR #574, opened 2023-01-15
Revert chain indentation in an argument list

belka-ew:
Fixes the indentation of arguments after a function chain (see the tests for examples).

This fix works only with `--single_indent=true`. The reason is that function chains aren't aligned properly without `--single_indent`. For example the following code:

```d
class C {
    void f()
    {
        if (true) {
            f(
                    array.map!(a => a.prop)
                        .array
                        .to!string,
                    __FILE__,
                    __LINE__);
        }
    }
}
```

is formatted as

```d
class C
{
    void f()
    {
        if (true)
        {
            f(
                    array.map!(a => a.prop)
                    .array
                    .to!string,
                    __FILE__,
                    __LINE__);
        }
    }
}
```

So I think it is a different problem.

### DIFF
--- a/tests/otbs/argument_chain_indent.d.ref
+++ b/tests/otbs/argument_chain_indent.d.ref
@@ -1,0 +1,10 @@
+class C {
+    void f() {
+        if (true) {
+            f(map.map.map.map.map.map.map.map.map.map.map.map.map.map.map.map
+                    .map.map.map.map.map.map, __FILE__, __FILE__, __FILE__, __FILE__, __FILE__,
+                __FILE__, __FILE__, __FILE__, __FILE__, __FILE__, __FILE__,
+                __FILE__, __FILE__, __FILE__, __FILE__, __FILE__, __LINE__);
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


tests/otbs/argument_chain_indent.d.ref
```diff
@@ -0,0 +1,10 @@
+class C {
+    void f() {
+        if (true) {
+            f(map.map.map.map.map.map.map.map.map.map.map.map.map.map.map.map
+                    .map.map.map.map.map.map, __FILE__, __FILE__, __FILE__, __FILE__, __FILE__,
```
WebFreak001, line 5: shouldn't there be a new line after the `.map,`? Or at least wrap after the 3rd `__FILE__` and not after the 5th (because the other lines below also wrap around that column)

WebFreak001 (approved):
looks good for what it fixes

tests/otbs/argument_chain_indent.d.ref
belka-ew: Yeah, it is how line breaks are currently added. It was the main reason we wanted to have `--keep-line-breaks`, so such things can be adjusted manually because in the complicated code it is difficult to calculate the right way to add line breaks (if there is the right way).

PR #585, opened 2023-07-26
No space after keywords

burner:
No more spaces after keywords like if,foreach,while,for, etc.


WebFreak001 (changes requested):
can you add a test case? Go into the `tests` folder, create a file like `space_after_keywords.d` with the various keyword statements described here (if, while, foreach, for, with), then run
```
echo '--space_after_keywords=false' > space_after_keywords.args
./gen_expected.sh space_after_keywords
```

confirm that the formatting is like what you expect and add it to this PR

burner:
done

WebFreak001 (approved):

burner:
awesome, thank you

PR #589, opened 2023-09-15
SAOC 2023: replace libdparse with dmd

snprajwal:
This PR tracks the commits for replacing libdparse with dmd. Unlike with other projects like D-Scanner, dfmt generates the AST only once (in the main function), and uses it across all 5 files. Since these files are tightly coupled together, making this change across multiple PRs is not possible as it will lead to a broken master branch. Instead, this PR will track every change in independent commits, and will be marked ready once dfmt+dmd works exactly how dfmt+libdparse previously did.

> *NOTE*: This PR is work in progress, and will most likely be broken for the foreseeable future.

cc @RazvanN7


dub.json
```diff
@@ -4,7 +4,10 @@
     "targetType": "autodetect",
     "license": "BSL-1.0",
     "dependencies": {
-      "libdparse": ">=0.19.2 <1.0.0"
+      "libdparse": ">=0.19.2 <1.0.0",
+      "dmd": {
+        "repository": "git+https://github.com/dlang/dmd.git"
```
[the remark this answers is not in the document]
snprajwal: Aha, thanks! I was using the git repo to ensure we were on the latest commit on master, but I suppose the registry dep should also work fine.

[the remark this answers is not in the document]
rikkimax: In the short term use a branch, but long term it'll need to be pinned to a version I'd expect.

You can always add-local for development, so not a problem to do it this way.

WebFreak001 (commented):
is it possible to keep the changes of formatter.d in there and not rename it to ast.d, so the diff can keep all the indentation related stuff if possible?

snprajwal:
> is it possible to keep the changes of formatter.d in there and not rename it to ast.d

I've not renamed `formatter.d`, I've just removed the code that will not be used anymore. `ast.d` has been introduced as a new file since it deals with walking the AST and doesn't fit into any of the existing files. `formatter.d` currently only contains the driver code for invoking the AST walk.

> so the diff can keep all the indentation related stuff if possible?

I'm not sure I understand what you're referring to as "indentation related"...

WebFreak001:
ah sorry I didn't actually check the diff too much there and was thinking about the layout inside libdparse, where there is also a formatter that works through the visitor pattern like this one does.

I think in the end the changes here will be a useful tool, being accurate with the compiler library, but I don't think the new code should go into this dfmt repository. I think it would be good to have this as a second repository in parallel that could support the same settings and reuse code such as the editorconfig parsing from dfmt, however the differences in how the two formatters are formatting code are too big to upgrade to with old code bases. Additionally new language features will require upgrade and maintenance each time, since otherwise code will be completely unable to parse/format, which with dfmt often just results in new features not looking good, but the rest still working as expected, so in the future it might still be a desired stable formatter that just doesn't get the latest and greatest features like this dmd replacement does.

Instead I think this new dmd-based formatter is something we can introduce slowly to new projects once it's ready. It fundamentally works differently to dfmt since it's actually rewriting the code based on the AST instead of just inserting/removing whitespace like what dfmt does. This of course also adds new opportunities, like being able to reorder imports alphabetically while formatting and adding/removing braces / shortened methods to be consistent with project settings - features which would never make it into dfmt with its current architecture.

For now I think it's good to keep it in this PR here, just for people to also look at it, but eventually we will probably want a new repo for it, for example calling it dmdfmt.

RazvanN7:
@WebFreak001 Our aim is to have the same functionality as the current dfmt, so there shouldn't be any breaking changes once this project is finished. However, I don't really understand how using dmd as a lib requires more upgrade & maintenance time as opposed to the status quo. Let's say a new feature is added in dmd. If you want to upgrade dfmt right now, you need to add some code to libdparse and then potentially add some code to dfmt to support the new case. With dmd as a lib, you will most likely need to update to a newer version and I suspect that in most cases you won't need to add anything new to the formatter or you would just need to update the dfmt code that uses the tree a bit. All in all, I expect the changes to require less time than what you need to do now.

Now, regarding old code bases: dfmt gets updates from time to time. There is nothing that forces you to use the latest version. If you have an old code base, then it makes sense to use an old compiler and an old version of the formatter. If you are using a new code, then you need a newer compiler and a newer formatter. The versioning system in dub (or manual versioning) should take care of this.

My personal opinion is that we should just upgrade dfmt and make sure that people can select the version proper version of the formatter that is in sync with the compiler version that they are using. However, if you fear breaking changes, one other way we can do it is to just create a branch with dmddfmt that goes in parallel with the old formatter. That way, people can select the branch that they are interested in. Having 2 separate repos that essentially do the same thing is not something that I am attracted to - we will need to always backport dfmt-frontend additions just to keep in sync.

maxhaton:
Is this working well enough to start getting stuff into the main tree?

maxhaton:
@RazvanN7 I'd like to try getting the tools into the dmd repo at some point.

snprajwal:
> Is this working well enough to start getting stuff into the main tree?

It largely works, but there's still some work left. It does not retain comments at the moment, and also does not allow all the configurations that dfmt supports. I'm working on hitting feature parity. I guess it's @RazvanN7's call to make about how this will be merged (into a different branch/a new repo/something else)

WebFreak001:
I think retaining comments is a very important issue to solve first before getting it to users, extending the feature support can be done later as well

snprajwal:
> I think retaining comments is a very important issue to solve first before getting it to users, extending the feature support can be done later as well

Agreed, I've been working on it across the last couple of days. It requires some changes in DMD that I'll have to make before adding support into dmdfmt for it, so it'll take some time.

snprajwal:
Here's the situation right now:

- DMD lexes comments, but does not store normal comments. It creates a `TOK.comment` token and moves on with the lexing (see [this line](https://github.com/dlang/dmd/blob/75869864da01e229807748a446c5738b313ededf/compiler/src/dmd/lexer.d#L699). It looks like it doesn't even create the token if we're using DMD as a library. It only stores the comment text if it is a doc comment (`///`, `/** **/`, or `/++ ++/`) -> see [this line](https://github.com/dlang/dmd/blob/75869864da01e229807748a446c5738b313ededf/compiler/src/dmd/lexer.d#L755).
- The doc comments stored in the previous step get added to the respective lang item nodes in the AST during parsing in ~78 places in [parse.d](https://github.com/dlang/dmd/blob/75869864da01e229807748a446c5738b313ededf/compiler/src/dmd/parse.d)

One way to store comments is to just change the lexing logic to store all types of comments rather than just doc comments. I don't think this is a feasible solution since it'll require a lot of changes across the rest of the compiler, especially in the DDoc generation logic. It also bloats the size of the AST during parsing with information that is effectively useless to the compiler.

Another way is to introduce a new field into `struct Token`, something like `rawComment`. It will store the raw comment strings that the lexer encounters, including the prefix and delimiter (`//`, `/* */`, etc). This will again require some change to be made to the parser to propagate this info into the AST nodes, but should not break anything that is currently present. This will allow us to simply dump the raw comment string into the output stream during formatting. I'm still not sure if this is the way to go, or if there is an even easier solution that I'm not able to think of.

The objectively "correct" way to solve this would be to introduce a new stage in the parsing phase where we build a concrete syntax tree (CST). This is the AST enriched with additional information including whitespaces, line breaks, comments, offset in the source file, etc. This CST can be used in the formatter to generate better context-aware output. The compiler will just have another pass that strips all of this info from the CST and reduces it to the current AST (before the semantic passes are run). Again, I'm not sure if this is the way to go, owing the the complexity of implementing this, potential of breaking the existing compiler pipeline, and adding additional overhead into the parsing phase.

WebFreak001:
in libdparse we solved this by Token having leading and trailing Trivia[] attached to it. Trivia is just tokens of type whitespace and comment (and some other skipped things)

in libdparse each AST node has an array of tokens it was constructed from, so you can eventually go all the way down to the trivia.

This is copied from Roslyn's design and makes it possibly to do a 1:1 reconstruction of the input code from the AST, even after modifying parts of it.

snprajwal:
I discussed this with Razvan, and we both feel that the correct way for us to go ahead would be to store this data correctly during the parsing stage and provide it through the AST, similar to having a `trivia[]` field. I'm working on adding this into DMD, and will open a PR once I make sure it works with dfmt.

On a tangent, with regards to the question about merging the work in this PR, I don't think we've reached a consensus yet on whether it should go into a separate `dmdfmt` project repo, or a separate branch in this repo.

WebFreak001:
since the functionality is so drastically different from how dfmt works I would suggest creating a separate project. Or alternatively it might be better to just have both the libdparse and dmd based formatters in dfmt, but be able to switch between them using a CLI switch like `--engine=dmd`, since it going to support the dfmt formatting flags as well.

RazvanN7:
@WebFreak001 I brought the issue of merging dmdlib with dfmt at our Foundation Monthly Meeting and the general consensus was to integrate this work into the mainline dfmt and have dfmt be part of the release archive. That way dfmt is going to be tied with a compiler version and everytime you update the compiler you also get the latest version of dfmt. How does that sound?

maxhaton:
Put it in the compiler tree!

WebFreak001:
as long as we keep both engines (libdparse and dmd) around for a while that sounds like a good idea, it's just that we will still want to support the old dfmt users for a while especially while early issues in the dmd issues are still being sorted out.

We don't want to just introduce completely different formatting behavior onto full projects that have relied on dfmt for a long time and are suddenly getting formatted differently (it will just make the git history of all these projects a mess as well)

src/dfmt/ast.d
```diff
@@ -0,0 +1,3926 @@
...
+import dmd.tokens;
+import dmd.parse;
+import dmd.mtype;
+import dmd.identifier;
+import dmd.hdrgen;
+import dmd.init;
+import dmd.rootobject;
+import dmd.target;
+import dmd.root.string : toDString;
+import dmd.optimize : optimize;
```
maxhaton, line 15: This is breaking CI for some reason.

maxhaton, line 15:
```suggestion

```
looks like it can be zapped

README.md
```diff
@@ -16,9 +16,9 @@ when using the **--inplace** option.
 
 ### Building from source using Make
 * Clone the repository
-* Run ```git submodule update --init --recursive``` in the dfmt directory
-* To compile with DMD, run ```make``` in the dfmt directory. To compile with
-  LDC, run ```make ldc``` instead. The generated binary will be placed in ```dfmt/bin/```.
+* Run `git submodule update --init --recursive` in the dfmt directory
```
maxhaton, line 19: Move these changes into a different PR if this is intentional.

.editorconfig
```diff
@@ -14,17 +14,5 @@ indent_size = 4
...
-dfmt_space_after_cast = true
-dfmt_space_after_keywords = true
-dfmt_split_operator_at_line_end = false
-end_of_line = lf
-indent_size = 4
-indent_style = space
-insert_final_newline = true
-max_line_length = 120
-tab_width = 8
+dfmt_single_template_constraint_indent = true
```
maxhaton, line 17: Why are these changes here? Intentional?

makefile
```diff
@@ -1,8 +1,36 @@
...
+
+DMD_PARSER_SRC := \
+	dmd/compiler/src/dmd/astbase.d \
+	dmd/compiler/src/dmd/parse.d \
+	dmd/compiler/src/dmd/parsetimevisitor.d \
+	dmd/compiler/src/dmd/transitivevisitor.d \
+	dmd/compiler/src/dmd/permissivevisitor.d \
+	dmd/compiler/src/dmd/strictvisitor.d \
+	dmd/compiler/src/dmd/astenums.d \
+	$(DMD_LEXER_SRC)
```
maxhaton, line 27: Might `dmd -i` be useful here? This makes you quite brittle towards upstream changes.

src/dfmt/formatter.d
```diff
@@ -1,21 +1,25 @@
...
 // Distributed under the Boost Software License, Version 1.0.
 //    (See accompanying file LICENSE.txt or copy at
 //          http://www.boost.org/LICENSE_1_0.txt)
 
 module dfmt.formatter;
 
-import dparse.lexer;
-import dparse.parser;
-import dparse.rollback_allocator;
-import dfmt.ast_info;
```
maxhaton, line 12: keep libdparse around for now, abstract your additions into some kind of API.

maxhaton (changes requested):

src/dfmt/ast.d
snprajwal: It's needed for calling `Expression.optimize`, which was moved out of `expression.d` into `optimize.d` in a recent PR. It breaks CI because I don't think there's been a release with the change yet, I compile dmdfmt with the master branch of dmd locally so I'm sometimes ahead of the CI pipeline. I'll update `dub.json` to use [v2.107.0-beta.1](https://github.com/dlang/dmd/releases/tag/v2.107.0-beta.1), that should fix it.

.editorconfig
snprajwal: Yes, the configs that were removed were redundant, because they're the defaults. The two lines that were added were to maintain consistency between multi-line formatting across the codebase (the default behaviour uses two tabs to indent, which is somewhat unidiomatic, imo). There's also a [discussion on the forum](https://forum.dlang.org/subscription-posts/lhugkccytdyaoybpczww) about whether to keep this knob or not on dmdfmt.

makefile
snprajwal: Yeah I'm not really sure about keeping these changes, because they refused to work once I started using `ASTCodegen`. I tried a bunch of things to fix it, but didn't manage to make it work. dmdfmt currently works with `dub build`, but the Makefile itself is pretty broken. Even with every single D file in DMD included, it would die during linking.

src/dfmt/formatter.d
snprajwal: You mean libdparse and dmd should be a switchable flag in dfmt?

.editorconfig
maxhaton: Its good practice (for good reason) to make PRs as small as possible. If you want to change this it should be a separate pull request for that reason.

src/dfmt/formatter.d
maxhaton: Yes, both for redundancy and because it will make your diff smaller. It's much easier to merge when you are adding rather than replacing.

src/dfmt/ast.d
maxhaton: and where

snprajwal: It's called [here](https://github.com/snprajwal/dfmt/blob/saoc-2023/src/dfmt/ast.d#L2211) and [here](https://github.com/snprajwal/dfmt/blob/saoc-2023/src/dfmt/ast.d#L2282). It was introduced to fix https://issues.dlang.org/show_bug.cgi?id=7375, and was taken from `hdrgen.d` in DMD. Now that I look closer at it, it's a compile time check, so shouldn't be needed in dfmt. I'll get rid of it.

PR #591, opened 2023-10-20
Fix issue 578

danielzuncke:
The infos are in the commit message, a bit odd is the position of the new array in `ASTInformation` (all other `size_t[]` arrays come before the `BraceIndentInfo` and `StructInitializerInfo` array). The idea is to prevent breaking any intializers someone may have if this is used as an import. I think that's unlikely though, let me know if I should order it by type.

Github tells me this merges 5 commits, but it should only be the last one **Fix issue 578**. I am not too familiar with github, if I am doing it wrong please let me know.

### DIFF
--- a/tests/allman/issue0578.d.ref
+++ b/tests/allman/issue0578.d.ref
@@ -4,40 +4,5 @@ void f()
     auto a = [true ? 1 : 0];
     auto aa1 = [0: true ? 1 : 0];
     auto aa2 = [0: true ? (false ? 1 : 2) : 3];
-
-    auto aa3 = [0: true ? false ? 1: 2 : 3];
-    /+
-    READ IF THIS TEST FAILS
-
-    Bug in dparse:
-    (Formatting before fix issue 578)
-    int[int] aa3 = [0: true ? false ? 1: 2: 3];
-                                       ^
-
-    EXPLANATION:
-    The marked colon is not is not recognized as a TernaryExpression by
-    dparse:
-    If you write a `writeln(ternaryExpression.colon.index)` in the overloaded
-    ASTInformation visit function, which should get called once for every
-    encountered ternary colon, only the second index is printed:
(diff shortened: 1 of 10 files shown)


tests/allman/issue0578.d.ref
```diff
@@ -0,0 +1,43 @@
+void f()
+{
+    auto t = true ? 1 : 0;
+    auto a = [true ? 1 : 0];
+    auto aa1 = [0: true ? 1 : 0];
+    auto aa2 = [0: true ? (false ? 1 : 2) : 3];
+
+    auto aa3 = [0: true ? false ? 1: 2 : 3];
```
WebFreak001, line 8: this seems to be a part of the issue you said would be fixed in the title, no?

danielzuncke: Yes, that's a bug in libdparse(*1), when that is fixed it will format correctly.
I added that test so when libdparse is fixed and this fails, someone can confirm that the new formatting is correct, adjust the test and remove the long explanation comment within the test.

\*1 From what I can tell libdparse doesn't recognise nested ternary expressions so the inner one doesn't get parsed by `visit(TernaryExpression ...)`. That doesn't matter for line 6 where when testing the associated array condition for the colon of the inner ternary expression, the `indents.topIs(tok!"]")` matches a parenthesis instead of a `]`.
I have opened [this issue](https://github.com/dlang-community/libdparse/issues/507#issue-1955324369) in libdparse.

danielzuncke: My bad I misunderstood libdparse, I fix it

WebFreak001 (approved):

WebFreak001:
please don't work on the master branch locally anymore, create branches for your fixes. You will have problems updating your master branch with upstream now unless you force push or recreate your fork

Issue #586, opened 2023-08-01
named function arguments


khvorov45:
Before formatting
```d
void temp(int v1, int v2) {}

void main() {
    temp(v1: 1, v2: 2);
    temp(
        v1: 1, 
        v2: 2,
    );
}
```
After formatting
```d
void temp(int v1, int v2) {
}

void main() {
    temp(v1 : 1, v2:
        2);
    temp(
v1 : 1,
v2:
        2,
    );
}
```

danielzuncke:
Would you like to have the option to choose between `temp(v1: 1, v2: 2)` and `temp(v1 : 1, v2 : 2)` like with `--space_before_aa_colon=false` or not?

khvorov45:
I'd like the option, sure.

WebFreak001:
makes sense for consistency and formatting code is likely to be the same as AA as well

redthing1:
I'm still having this issue in vscode, is my dfmt too old or something? My extension is up to date.

WebFreak001:
did you enable beta/nightly branch for serve-d?

Try upgrading to serve-d nightly by setting the `d.servedReleaseChannel` setting in vscode (Ctrl-, for User Settings) to `nightly` - afterwards, reload the window with `Ctrl-Shift-P` -> `Reload Window` and let it automatically download the update.

FeepingCreature:
I'm still seeing this issue with dfmt master.

Before formatting:
```
void test() {
    return Struct(
        foo: field.foo,
        bar: field.bar,
        baz: field.baz);
}
```
After formatting:
```
void test()
{
    return Struct(
foo: field.foo,
bar: field.bar,
baz: field.baz);
}
```

redthing1:
I am seeing the same thing as well, with the master branch.

PR #593, opened 2023-10-26
Fix issue #586

danielzuncke:
Feature from this implementation: handles struct initialisers with named arguments the same way as named function arguments (removing this might be difficult, also I think this is probably good for consistency):
```d
struct S { int i; }
void f(int i) {}
void main() {
    // both become this
    f(i: 1); auto s = S(i: 1);
    // or this
    f(i : 1); auto s = S(i : 1);
}
```

The multiline example from issue 586 is rewritten to one line if `--keep_line_breaks=false` which is not an error of the `:` formatting and / or expected behaviour (?) since the combined line is still rather short.
If `--keep_line_breaks=true` the argument indentation is wrong: but this shouldn't be related to `:` formatting and probably requires its own issue (if there isn't one yet):
```d
void main()
{
    f(
        v1: 1,
        v2: 2,
    );
}
```
becomes (with `--keep_line_breaks=true`)
```d
void main()
{
    f(
v1: 1,
v2: 2,
    );
}
```

\
There are no examples for
* comments around the function argument `f(/+ abc +/ i /+ def +/: 1);` since issue 565 messes with the result (the `:` part works as expected but tests will fail if 565 gets fixed).
* multiline anon func examples or similar multiline `{}` blocks inside the function call due to issues with indentation afterwards (probably related to issue 545).

### DIFF
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -461,31 +461,22 @@ final class FormatVisitor : ASTVisitor
             return;
         }
 
-        /+
-        Items are function arguments: f(<item>, <item>);
-        Iterate them and check if they are named arguments: tok!":" belongs to a
-        named argument if it is preceeded by one tok!"identifier" (+ any number
-        of comments):
-        +/
         foreach (item; functionCall.arguments.namedArgumentList.items)
         {
-            // Set to true after first tok!"identifier".
-            auto foundIdentifier = false;
+            // Do nothing if not a named argument.
+            if (item.name == tok!"")
+            {
+                continue;
+            }
 
+            // Find first colon if named argument.
             foreach (t; item.tokens)
             {
-                if (t.type == tok!"identifier" && !foundIdentifier)
-                {
-                    foundIdentifier = true;
-                    continue;
-                }
-
-                if (t.type == tok!":" && foundIdentifier)
+                if (t.type == tok!":")
                 {
                     astInformation.namedArgumentColonLocations ~= t.index;
+                    break;
                 }
-
-                break;
             }
         }
 
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -474,11 +474,6 @@ final class FormatVisitor : ASTVisitor
 
             foreach (t; item.tokens)
             {
-                if (t.type == tok!"comment")
-                {
-                    continue;
-                }
-
                 if (t.type == tok!"identifier" && !foundIdentifier)
                 {
                     foundIdentifier = true;
--- a/src/dfmt/ast_info.d
+++ b/src/dfmt/ast_info.d
@@ -64,6 +64,7 @@ struct ASTInformation
         sort(ufcsHintLocations);
         ufcsHintLocations = ufcsHintLocations.uniq().array();
         sort(ternaryColonLocations);
+        sort(namedArgumentColonLocations);
     }
 
     /// Locations of end braces for struct bodies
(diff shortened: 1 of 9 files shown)


src/dfmt/ast_info.d
```diff
@@ -448,6 +452,51 @@ final class FormatVisitor : ASTVisitor
...
+
+        /+
+        Items are function arguments: f(<item>, <item>);
+        Iterate them and check if they are named arguments: tok!":" belongs to a
+        named argument if it is preceeded by one tok!"identifier" (+ any number
+        of comments):
+        +/
+        foreach (item; functionCall.arguments.namedArgumentList.items)
+        {
+            // Set to true after first tok!"identifier".
```
WebFreak001, line 472: this has a bug with e.g. `foo([a: b])` inserting that colon, since it just goes over the tokens. Make sure you check if the named argument is even a named argument and not a regular argument, since the grammar for named argument list is:

```
     * $(GRAMMAR $(RULEDEF namedArgument):
     *     ($(RULE identifer) $(LITERAL ':'))? $(RULE assignExpression)
     *     ;)
```

so check for `if (item.identifier != tok!"")` before looking for a colon

(I just realized the picked name is not great, but this was just added since the AST closely resembles the grammar which was changed like this in the spec)

src/dfmt/ast_info.d
```diff
@@ -448,6 +452,51 @@ final class FormatVisitor : ASTVisitor
...
+            auto foundIdentifier = false;
+
+            foreach (t; item.tokens)
+            {
+                if (t.type == tok!"comment")
```
WebFreak001, line 477: comment tokens are not part of the `tokens` array in the AST, they are only used in trivia tokens (leadingTriviaTokens, trailingTriviaTokens) and are attached to the regular tokens, so you don't have to check for them

WebFreak001 (changes requested):

src/dfmt/ast_info.d
danielzuncke: Thanks for the quick feedback, I will change the identifier check to `if (item.identifier != tok!"")`. The aa example works correctly (I think all nested expressions should), but using the information from libdparse is ofc better.

I didn't know `item.identifier` exists, where can I look these things up? So far I went with the files that code-d finds when using "go to definition". The libdparse that I get with dub when building has this class:
```d
final class NamedArgument : BaseNode
{
    override void accept(ASTVisitor visitor) const
    {
        mixin (visitIfNotNull!(name, assignExpression));
    }
    mixin OpEquals;
    /** */ Token name;
    /** */ ExpressionNode assignExpression;
    /** */ size_t startLocation;
    /** */ size_t endLocation;
}
```
I did find this http://libdparse.dlang.io/grammar.html#namedArgument , where `identifier` exists but not for example `startLocation`, can I expect that members listed on the website exist additionally to what I can see in the class definition?

src/dfmt/ast_info.d
danielzuncke: Thanks for the info, I will remove the comment check.

src/dfmt/ast_info.d
danielzuncke: I wrote too soon, `item.identifier` isn't found, maybe I am using it wrong?
```d
foreach (item; functionCall.arguments.namedArgumentList.items)
{
    if (item.identifier == tok!"")
    {
        continue;
    }
    // ...
```
produces
```
src/dfmt/ast_info.d(472,21): Error: no property `identifier` for `item` of type `const(dparse.ast.NamedArgument)`
```

WebFreak001: ah sorry, it's `.name`

Yeah the ast.d file contains all the things, and to really understand where they come from in the grammar, you look at parser.d or the auto-generated docs on https://libdparse.dlang.io/grammar.html as well as possibly also the implementation

WebFreak001 (approved):

PR #600, opened 2024-05-02
makefile: small fixes

the-horo:


WebFreak001:
I don't know much about makefiles, anyone else willing to review this or who cares about the makefiles? Otherwise we can just merge this in

makefile
```diff
@@ -40,7 +39,7 @@ test: debug
 	cd tests && ./test.d
 
 bin/dfmt-test: bin/githash.txt $(SRC)
-	$(DC) $(DMD_TEST_FLAGS) $^ -of$@
+	$(DC) $(DMD_TEST_FLAGS) $(filter %.d,$^) -of$@
```
jtbx, line 42: Here `bin/githash.txt` is supplied as a dependency which will get passed to the D compiler when using `$^`, and the call to filter (also used in the bin/dfmt target) will make sure only .d files are passed.

jtbx (approved):

PR #612, opened 2025-09-25
Fix issue 586: Avoid interpreting named args as labels when indenting newline

FeepingCreature:


FeepingCreature:
does now

WebFreak001 (approved):
this whole thing makes me think it would be way easier and less error-prone to just check if we are within a `(...)` context and not within a `{...}` context

PR #613, opened 2025-10-06
Issue 586 followup: Tag the constructor arguments in new-anonymous-class calls as named arguments.

FeepingCreature:
Forgot that case.


FeepingCreature:
there's one more where it goes wrong but I don't remember what it was, I'll fix it if I find it again

WebFreak001 (approved):

PR #614, opened 2026-02-17
Add --force_curly_braces option to insert braces around single-statem…

burner:
…ent bodies.

When enabled, the preprocessor inserts synthetic `{` and `}` tokens around braceless bodies for all block header constructs: if, else, for, foreach, while, do, scope, synchronized, with, catch, debug, version, finally, and their combinations (static if, static foreach, etc.). Implementation uses token stream preprocessing so the formatter's existing brace-handling logic handles style and indentation naturally. Synthetic tokens use size_t.max-based indices to avoid collisions with AST info lookups. Handles compound constructs correctly:
- try/catch/finally chains as single statements
- Nested braceless blocks (each level gets its own braces)
- Mixed braced/braceless siblings (only braceless branches get braces)
- else if/else version/else static if chains (no double-wrapping)
- debug/version with and without parens
- Respects // dfmt off / // dfmt on regions

### DIFF
--- a/tests/force_curly_braces.d
+++ b/tests/force_curly_braces.d
@@ -244,3 +244,15 @@ unittest
     if (imin.value > 0x10FFFFUL) // One
         imin.value = 0x10FFFFUL; // Two
 }
+
+// 27. Complex version/else/debug/else chain
+unittest
+{
+    version (Windows)
+        __locale_decpoint = save;
+    else version (Foo)
+        __locale_decpoint = save;
+    else
+        debug __x = 3;
+        else foobar();
+}
--- a/tests/force_curly_braces.d
+++ b/tests/force_curly_braces.d
@@ -1,0 +1,186 @@
+// Test: force_curly_braces option
+// Tests various braceless constructs that should get braces added
+
+// 1. Simple if without braces
+int test1(int x) {
+    if (x > 0)
+        return 1;
+    return 0;
+}
+
+// 2. if/else without braces
+int test2(int x) {
+    if (x > 0)
+        return 1;
+    else
+        return 0;
+}
+
+// 3. Nested if without braces
+void test3(int a, int b) {
+    if (a)
+        if (b)
+            doSomething();
+}
+
+// 4. if/else if/else chain
+int test4(int x) {
+    if (x > 0)
+        return 1;
+    else if (x < 0)
+        return -1;
+    else
+        return 0;
+}
+
+// 5. for loop without braces
+void test5() {
+    for (int i = 0; i < 10; i++)
+        doSomething();
+}
+
+// 6. foreach without braces
+void test6(int[] arr) {
+    foreach (x; arr)
+        doSomething();
+}
+
+// 7. while without braces
+void test7() {
+    while (condition())
+        doSomething();
+}
+
+// 8. Mixed: if braced, else not
+int test8(int x) {
+    if (x > 0) {
+        return 1;
+    } else
+        return 0;
+}
+
+// 9. Mixed: if not braced, else braced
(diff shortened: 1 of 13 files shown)


tests/force_curly_braces.d
```diff
@@ -0,0 +1,186 @@
...
+// 13. do-while without braces
+void test13() {
+    do
+        doSomething();
+    while (condition());
+}
+
+// 14. scope(exit) without braces
+void test14() {
+    scope(exit)
```
WebFreak001: can you also add a test to check that attributes like `scope T!() x = 5;` are not somehow getting curly braces added?

Also did you check how the tests acted if you enabled this new option on all the existing tests? Might be a good sanity check for any bugs in the implementation since I'd think most (not all) tests use braces everywhere already

burner: > Also did you check how the tests acted if you enabled this new option on all the existing tests? Might be a good sanity check for any bugs in the implementation since I'd think most (not all) tests use braces everywhere already

I did not, I'm not sure how.

burner: I added the scope variables and there seems to be no issue.

WebFreak001:
I just edited the code locally to enable your flag by default (`dfmt_force_curly_braces = OptionalBoolean.t`) and then reran the tests and checked some outputs:

Some bugs I encountered:
```
void foo() {
	string command;
	version (Posix) command ~= " 2> /dev/null 1> /dev/null";

	version (Posix) command ~= " 2> /dev/null 1> /dev/null";

	unittest
	{
		version (Posix) command ~= " 2> /dev/null 1> /dev/null";
	}
}
```
results in weird whitespace - I would like if the empty line could be after the closing brace instead:

```
void foo()
{
    string command;
    version (Posix)
    {
        command ~= " 2> /dev/null 1> /dev/null";

    }
    version (Posix)
    {
        command ~= " 2> /dev/null 1> /dev/null";

    }
    unittest
    {
        version (Posix)
        {
            command ~= " 2> /dev/null 1> /dev/null";
        }
    }
}
```

---

and it would be nice if you could make comments work, but not a dealbreaker:
```
unittest
{
	if (imin.value > 0x10FFFFUL) // ??
		imin.value = 0x10FFFFUL; // ??
}
```
->
```
unittest
{
    if (imin.value > 0x10FFFFUL) // ??
    {
        imin.value = 0x10FFFFUL;
    } // ??
}
```

---

The whitespace here is a bit broken, but the braces are fine imo:
```
unittest
{
	version (Windows) __locale_decpoint = save;
	else version (Foo) __locale_decpoint = save;
	else debug __x = 3;
	else foobar();
}
```
->
```

unittest
{
    version (Windows)
    {
        __locale_decpoint = save;
    }
    else version (Foo)
    {
        __locale_decpoint = save;
    }
    else
    {
        debug
        {
            __x = 3;
        }
    else
        {
            foobar();
        }
    }
}
```

---

It doesn't support debug / else debug / else blocks yet:
```
void main() {
debug (Foo) foo();
else debug (Bar) bar();
else baz();
}
```
->
```
void main()
{
    debug (Foo)
    {
        foo();
    }
    else
    {
        debug (Bar)
        {
            bar();
        }
        else
        {
            baz();
        }
    }
}
```

---

Perhaps `with (...) switch (...)` blocks should not be nested since they are kind of a paradigm to write together, but I'm not particularly set on that idea:
```
unittest
{
	with (SomeEnum) final switch (value)
	{
	case a:
		aa();
		break;
	case b:
		bb();
		break;
	}
}
```
should stay and not get the `{}` for the `with (SomeEnum)`

Issue #601, opened 2024-05-15
`@trusted` and `return` attributes get sandwiched


dkorpel:
```D
struct S
{
    void f() @trusted return
    {
    }
}
```

Gets formatted to:

```D
struct S
{
    void f() @trustedreturn
    {
    }
}
```

Turning the two attributes into one erroneous attribute.

PR #615, opened 2026-04-12
Fix missing space before return as function attribute

v0lv0:
https://github.com/dlang-community/dfmt/issues/601


PetarKirov (approved):
The fix looks solid and correctly handles the missing space before `return` when preceded by an attribute identifier, mimicking how `formatKeyword()` processes standard keywords. Tests pass successfully. 

*(Note: I also noticed the `do` keyword suffers from the exact same issue when used as a contract after an attribute, e.g., `void f() @safe do {}` formats to `@safedo`. This can be fixed in a future PR by adding `tok!"identifier"` to the `peekBackIsFunctionDeclarationEnding()` check!)*

